PR #640, opened 2023-01-31
IEC61850 Goose

PJungkamp:
A new node type able to subscribe to IEC61850 GOOSE events.

# Progress

- [x] Documentation

- [ ] `GooseNode` VLAN tags  
Neither subscriber nor publisher is currently capable of properly using GOOSE VLAN tags.

- [x] `GooseNode` Subscriber: change trigger
The subscriber code handling 'change' as a trigger for samples will currently miss changes if the message with sequence number 0 ist lost.

- [x] `GooseNode` Subscriber: handle unsupported MmsValues
The subscriber should not throw an exception when an unsupported type is received (e.g. VisibleString or OctetString).

- [x] `GooseNode` Publisher: change detection
Every written sample is currently indicating a state change in the state number. The `_write` code should save and compare against the last written sample.

- [x] `GooseNode` Publisher: burst on change
A common approach to ensure that change messages are not missed is to send a burst of messages on change.

- [x] `GooseNode` Publisher: periodic resend
Spawn a thread which periodically resends the last values in the background


stv0g:
Hi @PJungkamp, thanks for the contribution :)

I converted this PR to a draft. Please let me know once its ready for review.

etc/examples/nodes/iec61850-goose.conf
```diff
@@ -0,0 +1,66 @@
+
+nodes = {
+	goose = {
+		type = "iec61850-goose"
+
+		in = {
+			# ethernet interface to listen on
```
stv0g, line 7: Please start comments with uppercase letter

etc/examples/nodes/iec61850-goose.conf
```diff
@@ -0,0 +1,66 @@
...
+					trigger = "change"
+				}
+			}
+
+			# mapping from goose events to signals
+			signals = (
+				{
+					name = "ABB_relay_state"
+					type = "boolean"
+					# mandatory MmsType specification
```
stv0g, line 38: Please add an empty line between the options for readability

etc/examples/nodes/iec61850-goose.conf
```diff
@@ -0,0 +1,66 @@
...
+					mms_type = "bit-string"
+					subscriber = "relay"
+					index = 1
+				}
+			)
+		}
+	}
+}
+
+paths = (
```
stv0g, line 57: The node example files usually come without a `paths` section.

include/villas/nodes/iec61850_goose.hpp
```diff
@@ -0,0 +1,184 @@
...
+		SignalType signal_type;
+		char const *name;
+		std::optional<MmsType> mms_type;
+		Meta default_meta;
+	};
+
+	inline static std::array const descriptors {
+		Descriptor { SignalType::INVALID,	"invalid" },
+
+		/* boolean signals */
```
stv0g, line 71: Please use C++ style comments `//` rather than `/* */`

Also start comments with upper case letter

include/villas/nodes/iec61850_goose.hpp
```diff
@@ -0,0 +1,184 @@
...
+#include <villas/signal.hpp>
+#include <libiec61850/goose_receiver.h>
+#include <libiec61850/goose_subscriber.h>
+
+namespace villas {
+namespace node {
+namespace iec61850 {
+
+// A GooseValue is a SignalData value with attached Metadata for the MmsType and SignalType
+class GooseValue {
```
stv0g, line 29: I would propose to name this `GooseSignal` as this is more aligned with the `Signal` type in VILLASnode.

Or do you have a specific reason for naming it value?

include/villas/nodes/iec61850_goose.hpp
```diff
@@ -0,0 +1,184 @@
...
+
+private:
+	struct Descriptor {
+		SignalType signal_type;
+		char const *name;
+		std::optional<MmsType> mms_type;
+		Meta default_meta;
+	};
+
+	inline static std::array const descriptors {
```
stv0g, line 68: I already started to build a descriptor table for IEC61850 SV here:

https://github.com/VILLASframework/node/blob/37c9efad92de55f62bd74af97eb755e9b350390e/lib/nodes/iec61850.cpp#L27

I would propose to use / extend the same table for GOOSE ans SV.

lib/nodes/iec61850_goose.cpp
```diff
@@ -0,0 +1,577 @@
...
+#include <villas/super_node.hpp>
+#include <villas/exceptions.hpp>
+
+using namespace villas;
+using namespace villas::node;
+using namespace villas::utils;
+using namespace villas::node::iec61850;
+using namespace std::literals::chrono_literals;
+
+static std::optional<std::array<uint8_t, 6>> stringToMac(char *mac_string) {
```
stv0g, line 22: Please move `{` to new line

lib/nodes/iec61850_goose.cpp
```diff
@@ -0,0 +1,577 @@
...
+{
+	return descriptor->name;
+}
+
+SignalType GooseValue::signalType() const
+{
+	return descriptor->signal_type;
+}
+
+GooseValue GooseValue::fromMmsValue(MmsValue *mms_value) {
```
stv0g, line 53: Please move `{` to new line

lib/nodes/iec61850_goose.cpp
```diff
@@ -0,0 +1,577 @@
...
+	auto mms_type = MmsValue_getType(mms_value);
+	auto descriptor = lookupMmsType(mms_type);
+	SignalData data;
+
+	if (!descriptor) return {};
```
stv0g, line 58: `{}` is not needed here?

lib/nodes/iec61850_goose.cpp
```diff
@@ -0,0 +1,577 @@
...
+	};
+
+	auto descriptor = std::find_if(begin(descriptors), end(descriptors), check);
+	if (descriptor != end(descriptors))
+		return &*descriptor;
+	else
+		return std::nullopt;
+}
+
+GooseValue::GooseValue()
```
stv0g, line 184:
```c++
GooseValue::GooseValue() :
     signal_data({}),
     descriptor(&descriptors[0])
{ }
```

lib/nodes/iec61850_goose.cpp
```diff
@@ -0,0 +1,577 @@
...
+: signal_data({})
+, descriptor(&descriptors[0])
+{
+}
+
+GooseValue::GooseValue(GooseValue::Descriptor const *descriptor, SignalData data)
```
stv0g, line 190: Same as above :)

lib/nodes/iec61850_goose.cpp
```diff
@@ -0,0 +1,577 @@
...
+, descriptor(descriptor)
+{
+}
+
+void GooseNode::onEvent(GooseSubscriber subscriber, GooseNode::InputEventContext &ctx) noexcept
+{
+	if (!GooseSubscriber_isValid(subscriber) || GooseSubscriber_needsCommission(subscriber))
+		return;
+
+	if (	ctx.subscriber_config.trigger == InputTrigger::CHANGE
```
stv0g, line 201: Please use the same indention as in the rest of the code. And sorry for beeing so picky about it...

lib/nodes/iec61850_goose.cpp
```diff
@@ -0,0 +1,577 @@
...
+}
+
+void GooseNode::pushSample(uint64_t timestamp) noexcept
+{
+	Sample *sample = sample_alloc(&input.pool);
+	if (!sample) {
+		logger->warn("Pool underrun");
+		return;
+	}
+
```
stv0g, line 233: Superfluous empty line

lib/nodes/iec61850_goose.cpp
```diff
@@ -0,0 +1,577 @@
...
+int GooseNode::parse(json_t *json, const uuid_t sn_uuid)
+{
+	int ret;
+	json_error_t err;
+
+	ret = Node::parse(json, sn_uuid);
+	if (ret)
+		return ret;
+
+	ret = _parse(json, &err);
```
stv0g, line 393: Please merge `_parse` into this function.

I dont see any other callsite of `_parse`. So I dont see a reason to factor it out to a dedicated function?

include/villas/nodes/iec61850_goose.hpp
```diff
@@ -0,0 +1,184 @@
...
+	union Meta {
+		MetaSize size;
+	} meta;
+
+	SignalData signal_data;
+
+private:
+	struct Descriptor {
+		SignalType signal_type;
+		char const *name;
```
stv0g, line 63: Lets use a `std::string` here. It will also allow to get rid of some the `strcmp()` lateron..

stv0g (changes requested):
Great work @PJungkamp 🥳 

I added a few small comments into the code. Here are also some more general ones:

- Have you checked to pass const pointers/references as function arguments whereever it is suitable?
- @windrad6 mentioned that there is quite some overlap of the GOOSE code with the possible future MMS implementation. Hence it would be nice to keep this in mind and maybe move some of the code to `iec61850.cpp` (but we can do this also later..)

etc/examples/nodes/iec61850-goose.conf
PJungkamp: Fixed

include/villas/nodes/iec61850_goose.hpp
PJungkamp: I originally had the GooseValue wrap a MmsValue instead of SignalData. I already thought about renaming to MmsSignal. Though GooseSignal makes more sense for now.

include/villas/nodes/iec61850_goose.hpp
PJungkamp: I'd like to merge the descriptors for Goose and SV. But I think this makes more sense to do while refactoring parts of the SV node as well. Though I should keep the stringified names of the goose and sv nodes in sync.

lib/nodes/iec61850_goose.cpp
PJungkamp: Fixed.

lib/nodes/iec61850_goose.cpp
PJungkamp: Fixed.

lib/nodes/iec61850_goose.cpp
PJungkamp: I wanted to only have one site where the `throw ConfigError(json, err, "node-config-node-iec61850-goose")` occures. This allowed all parse related functions to adhere to the `if (ret) return ret;` pattern to propagate json parsing errors.

include/villas/nodes/iec61850_goose.hpp
PJungkamp: Fixed.

PJungkamp:
Some notes:

-  `GooseSignal`  
I'm still somewhat unhappy with the GooseSignal as I'm still undecided on whether I want to allow default construction.
It also should probably reuse the type descriptors from the SV node.
I'd prefer refactoring to merge duplicate code between SV and Goose in a seperate MR.

- `GooseNode` Publisher    
I'm undicided on whether I want to support a mode where the node spins up a thread to resend the last sample for every ... seconds.
This seems like a rather complex feature to support and configure.

Thank you for the preliminary review!

lib/nodes/iec61850_goose.cpp
stv0g: The ` if (ret) return ret;` pattern is actually a relict from the older C code. I think we should migrate to use exceptions for configuration parsing.

It has also the advantage, that we can more clearly report where the error occurred (line number of JSON config).
Have a look at the other occurences of ConfigError. In many instances we also included an ID which we used to point the user directly to the documentation where the setting is documented.

lib/nodes/iec61850_goose.cpp
```diff
@@ -0,0 +1,873 @@
...
+
+	err = queue_signalled_close(&input.queue);
+
+	return Node::stop();
+}
+
+static char name[] = "iec61850-goose";
+static char description[] = "Subscribe to and publish GOOSE messages";
+static NodePlugin<
+	GooseNode,
```
stv0g, line 866: Hi while I like this way of formatting the template args, its not consistent with the rest of the code.

Would you mind to use the same style as in the other files?

I would be happy to accept a dedicated PR which changes the formatting to your new style project-wide.

stv0g:
Good progress @PJungkamp :) I am happy to see a new node-type coming to VILLASnode

stv0g (approved):
Thanks a lot :-)

PR #644, opened 2023-05-25
tools/villas: improve script behaviour for `-h` or `--help`

PJungkamp:
- Quote `${SUBTOOL}` in case it is set but empty
- Add `-e` flag to `grep` so `villas --help` does not show `grep --help`
- Add `-F` flag to `grep` so `villas "[ "` does not show a regexp error
- Don't exit with error when `-h` or `--help` was specified
- Add `-h` and `--help` to `Usage` in the help message.

### DIFF
--- a/tools/villas
+++ b/tools/villas
@@ -22,8 +22,8 @@ SUBTOOL=$1
 ARGS=${@:2}
 
 # Check if tool is available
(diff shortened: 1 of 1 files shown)


stv0g:
Hi @PJungkamp,

thanks a lot :) Seems like you are late up like me coding :D

PR #668, opened 2023-06-20
webrtc: Remove dead code

stv0g:


PJungkamp (approved):
I forgot those... Looks good. 👍

PR #670, opened 2023-06-20
Webrtc signaling fixes

stv0g:


PJungkamp:
Since the conceal value in sul is higher than the length of the back off table, lws will should try to reconnect indefinitely. Should we make the reconnection attempts exhausted message a hard exception?

stv0g:
I would keep this behaviour for now. WhatsApp the maximum interval between consecutive Retriever I the current Code?

PJungkamp:
https://github.com/VILLASframework/node/blob/master/include/villas/nodes/webrtc/signaling_client.hpp#L45

`2<<16` milliseconds, so about a minute.

stv0g:
Okay that's acceptable

PR #677, opened 2023-06-22
Disable warning as error for normal release builds

PJungkamp:


stv0g:
Hi @PJungkamp, please feel free to remove the draft status once its ready for review.

PJungkamp:
Will do. It would break the pipeline for master though, as the pool test keeps failing.

PJungkamp:
The pool unit test relied on global initialization order across translation units. This undefined behaviour caused the tests to fail in release builds.

PR #686, opened 2023-06-29
Refactor WebRTC signaling and Node UUID initialization

### LINKED ISSUE #684: No not pass super-node to UUID node-type implementations

stv0g:
Fixes #684


PJungkamp:
The FpgaNode constructor needs to be converted to use take a `uuid_t`.

PJungkamp (dismissed):
Looks fine

PJungkamp:
I'm wondering how the `LoopbackNode` was ever working. The `parse` call passed an `int` to the format args of `json_unpack` instead of the appropriate `int *`. The `int queuelen` is also always initialized as non-zero. This non-zero int is then interpreted as a pointer by `json_unpack`. How was the loopback.conf in the examples ever working when specifying the `queuelen` is pretty much a guaranteed segfault?

PJungkamp (approved):
I think this is now ready to merge. @stv0g could you sanity check the last 3 commits.
Could your also update the signaling server and the server repo before merging?

stv0g:
I believe we did not really overload `parse()` but in realiity only ever called `Node::parse()` before..

Thanks for finding this :)

PR #693, opened 2023-07-07
Allow minimal CMake build configuration

PJungkamp:
Enabling all available CMake options for packaging, testing and CI by default makes sense.
A full configuration ensures that end users of VILLAS are not missing component which they might expect from looking at the documentation.

But especially on constraint systems (e.g. embedded) a minimal configuration with explicit opt in for compile options may be desirable.

This PR adds the `WITH_DEFAULTS` option to build without all optional features, even those theoretically available on the building system. One can then enable nodes, tools and similar by **opt-in** (e.g. `-DWITH_<feature>=ON`).


stv0g:
LGTM :)

stv0g (approved):

PR #695, opened 2023-07-25
add CODEOWNERS

n-eiling:
Add codeowners. After merging we should activaer "Require review from Code Owners" in the github settings.


stv0g (approved):
Very cool :)

PR #696, opened 2023-08-22
New node-type: Modbus

PJungkamp:
An initial version for modbus support in VILLASnode.


lib/nodes/modbus.cpp
```diff
@@ -0,0 +1,915 @@
...
+{}
+
+SignalData RegisterMappingSingle::read(uint16_t const *registers, modbus_addr_t length) const
+{
+	SignalData data;
+
+	if (num_registers() != length)
+		throw RuntimeError { "reading from invalid register range" };
+
+	if (auto i2i = std::get_if<IntegerToInteger>(&conversion)) {
```
stv0g, line 171: Can we remove the `{}` for single statement branches?

lib/nodes/modbus.cpp
```diff
@@ -0,0 +1,915 @@
...
+
+	return data;
+}
+
+void RegisterMappingSingle::write(SignalData data, uint16_t *registers, modbus_addr_t length) const
+{
+	if (num_registers() != length)
+		throw RuntimeError { "writing to invalid register range" };
+
+	if (auto i2i = std::get_if<IntegerToInteger>(&conversion)) {
```
stv0g, line 191: Can we remove the `{}` for single statement branches?

stv0g (changes requested):
Looks verify good :) Just a few stylistic change requests :)

stv0g (approved):

PR #704, opened 2023-08-25
Don't truncate files in `file` node-type

### LINKED ISSUE #701: `file` node-type can't read files.
The `file` node-type always truncates the file unconditionally (fopen with "w+").
This has been broken for about 2 years.

Seems like [this](https://github.com/VILLASframework/node/commit/784e970bfe48cf75da7622ca4c4a13fb529cce59) commit changed the file mode from "a+" to "w+".

PJungkamp:
Fixes #701

### DIFF
--- a/lib/nodes/file.cpp
+++ b/lib/nodes/file.cpp
@@ -262,7 +262,7 @@ int villas::node::file_start(NodeCompat *n)
 	f->formatter->start(n->getInputSignals(false));
 
 	/* Open file */
-	f->stream_out = fopen(f->uri, "w+");
+	f->stream_out = fopen(f->uri, "a+");
 	if (!f->stream_out)
 		return -1;
 


stv0g:
Unfortunately, this broke the node-multiplex integration test here:
https://github.com/VILLASframework/node/blob/65cb2d612be341fd1896a843a17601b679f9264c/tests/integration/node-multiplexing.sh#L65

This test runs a loop and hence appends to the output file.

PJungkamp:
> Unfortunately, this broke the node-multiplex integration test here:
> 
> https://github.com/VILLASframework/node/blob/65cb2d612be341fd1896a843a17601b679f9264c/tests/integration/node-multiplexing.sh#L65
> 
> This test runs a loop and hence appends to the output file.

As it should according to the documentation. I'll adapt the test.

stv0g (approved):

stv0g:
Thanks :)

PR #705, opened 2023-08-25
Add hook to fix reordered samples

PJungkamp:


lib/hooks/reorder_ts.cpp
```diff
@@ -0,0 +1,168 @@
...
+				if (i != window.size() - 1) {
+					logger->warn("Fixing reordered Sample");
+					sample_dump(logger, sample);
+				}
+
+				if (window.size() == window_size) {
+					// The front sample will be returned.
+					Sample *window_sample = window.front();
+					// Move all elements before the index of insertion towards the front.
+					std::memmove(&window[0], &window[1], i * sizeof(decltype(window)::value_type));
```
stv0g, line 121: Is there a reason you are using `std::memmove` here? Would't `std::vector::insert()` be easier to understand?

stv0g (changes requested):

lib/hooks/reorder_ts.cpp
PJungkamp: Using `insert` with iterators into `this` is undefined behaviour. The `std::copy` and `std::copy_backward` algorithms might work, but seemed similar in verbosity.

lib/hooks/reorder_ts.cpp
```diff
@@ -0,0 +1,160 @@
...
+namespace node {
+
+class ReorderTsHook : public Hook {
+
+protected:
+	std::vector<Sample *> window;
+	std::size_t window_size;
+	Sample *buffer;
+
+	void swapSample(Sample *lhs, Sample *rhs) {
```
stv0g, line 28:
```suggestion
	void swapSample(Sample *lhs, Sample *rhs)
	{
```

stv0g (approved):

lib/hooks/reorder_ts.cpp
```diff
@@ -0,0 +1,161 @@
...
+			window.push_back(sample_clone(smp));
+
+			logger->debug("window.size={}/{}", window.size(), window_size);
+
+			return Hook::Reason::SKIP_SAMPLE;
+		}
+
+		for (std::size_t i = window.size() - 1;; i--) {
+			if (time_cmp(&smp->ts.origin, &window[i]->ts.origin) >= 0) {
+				if (i != window.size() - 1) {
```
stv0g, line 106: Please remove `{}` for single statements

stv0g (approved):
Looks good :) I think we can merge this now.

Can you address the last comment which I added and also add the missing OpenAPI spec for the hook config?

Thanks :)

PJungkamp:
- [x] Added OpenAPI specification.
- [x] Fixed `if` braces.

stv0g (approved):

PR #706, opened 2023-08-25
Fixes to deps.sh and netem support for modbus node-type

### LINKED ISSUE #698: Return TCP socket descriptor in `ModbusNode::getNetemFDs()`
Only in TCP mode as this will fail for the RTU mode.

@PJungkamp

stv0g:
Fixes #698, #697

### DIFF
--- a/lib/nodes/modbus.cpp
+++ b/lib/nodes/modbus.cpp
@@ -904,6 +904,17 @@ std::vector<int> ModbusNode::getPollFDs()
 	return { read_task.getFD() };
 }
 
+std::vector<int> ModbusNode::getNetemFDs()
+{
+	if (modbus_context != nullptr && std::holds_alternative<Tcp>(connection_settings))  {
+		return {
+			modbus_get_socket(modbus_context)
+		};
+	}
+
+	return { };
+}
+
 const std::string & ModbusNode::getDetails()
 {
 	if (details.empty()) {
--- a/lib/nodes/modbus.cpp
+++ b/lib/nodes/modbus.cpp
@@ -35,6 +35,11 @@
  * @license Apache 2.0
  *********************************************************************************/
 
+#include <atomic>
+#include <chrono>
+
+#include <fmt/format.h>
+
 #include <villas/node_compat.hpp>
 #include <villas/nodes/modbus.hpp>
 #include <villas/utils.hpp>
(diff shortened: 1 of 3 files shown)


lib/nodes/modbus.cpp
```diff
@@ -904,6 +904,17 @@ std::vector<int> ModbusNode::getPollFDs()
 	return { read_task.getFD() };
 }
 
+std::vector<int> ModbusNode::getNetemFDs()
+{
+	if (modbus_context != nullptr && std::get_if<Tcp>(&connection_settings) != nullptr)  {
```
PJungkamp, line 909: You should use [`std::holds_alternative`](https://en.cppreference.com/w/cpp/utility/variant/holds_alternative) if you're only interested in the current variant. `std::get_if` should only be used if you need access to the contents of the variant.

packaging/deps.sh
```diff
@@ -40,7 +40,8 @@ should_build() {
 
     if [[ -z "${DEPS_NONINTERACTIVE+x}" ]] && [[ -t 1 ]]; then
         echo
-        case "$(read -p "Do you wan't to install '${id}' into '${PREFIX}'? This is used ${use}. (y/N)")" in
+        read -p "Do you wan't to install '${id}' into '${PREFIX}'? This is used ${use}. (y/N) "
+        case "${REPLY}" in
```
PJungkamp, line 44: You should probably add a variable name to the read command as `REPLY` is parsed slightly differently than other variables.

See the bash `read` manpage: https://ss64.com/bash/read.html
> If no names are supplied, the line read is assigned to the variable $REPLY.

This means `REPLY` also includes all whitespace in the line, while an explicit name has the whitespace trimmed.

stv0g: Strange, the old approach didnt work for me.. My new one does. I also cant see that it appends a newline:

```bash
stv0g@GER-L-06044:/$ read
hallo
stv0g@GER-L-06044:/$ echo $REPLY
hallo
stv0g@GER-L-06044:/$ [[ $REPLY == "hallo" ]] && echo matches
matches
```

PJungkamp:
There is a segfault in the CI integration tests. Is that spurious? This does not seem to be caused by the netem fds from libmodbus.

stv0g:
Yes thats a spurious test failure in the queue.

PJungkamp (approved):

PR #718, opened 2023-09-29
python: Add digest-hook fifo format parsing

PJungkamp:


stv0g (approved):
Perfect 👍🏻 No comments, we can merge it as is

PR #722, opened 2024-02-08
Fix formatting using clang-format

stv0g:


fwege (approved):

lib/nodes/rtp.cpp
n-eiling, line 1: I think the order matters in this file. The CI fails because some types are unknown.

n-eiling (changes requested):

lib/nodes/rtp.cpp
stv0g: Thanks. I've locally disabled clang-format for this section of the file.

n-eiling (approved):

PR #724, opened 2024-02-17
Node fix multiple sources

pipeacosta:
This PR fixes configurations in which multiple paths are defined with the same source node

### DIFF
--- a/lib/nodes/loopback_internal.cpp
+++ b/lib/nodes/loopback_internal.cpp
@@ -23,13 +23,12 @@ InternalLoopbackNode::InternalLoopbackNode(Node *src, unsigned id, unsigned ql)
     : queuelen(ql), source(src) {
   auto name = fmt::format("{}.lo{}", src->getNameShort(), id);
 
-  uuid_t uuid;
   int ret =
       uuid::generateFromString(uuid, fmt::format("lo{}", id), src->getUuid());
   if (ret)
     throw RuntimeError("Failed to initialize UUID");
 
-  Node(uuid, name);
+  name_short = name;
 
   factory = &nf;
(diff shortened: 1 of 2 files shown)


stv0g:
Hi @pipeacosta,

did you see that the CI tests are failing? I think you need to rebase your changes to master.
Niklas has fixed the CI issue already there.

I am also wondering about the motivation for this change. Can you quickly elaborate by you do not call the parent constructor (`Node()`) anymore?

pipeacosta:
> Hi @pipeacosta,
> 
> did you see that the CI tests are failing? I think you need to rebase your changes to master. Niklas has fixed the CI issue already there.
> 
> I am also wondering about the motivation for this change. Can you quickly elaborate by you do not call the parent constructor (`Node()`) anymore?

Hi @stv0g,

Because the parent constructor gets called anyway once `InternalLoopbackNode` is instantiated. If no id is provided, the attribute `uuid` gets automatically generated using [uuid_generate_random](https://github.com/VILLASframework/node/blob/73ff061ca83886ab031f8f5e8cbea75560a2b1a9/lib/node.cpp#L58). I guess the problem uuid is defined again [here](https://github.com/VILLASframework/node/blob/73ff061ca83886ab031f8f5e8cbea75560a2b1a9/lib/nodes/loopback_internal.cpp#L26). Then, after realizing that the parent constructor does get called, what I did is just to assign the new value of `name_short`. In this way, I noticed that the segmentation fault error that arises when you set up multiple sources, like in the `node_multiple_sources.sh` integration test is solved. Hope this clarifies your doubts a bit ...

stv0g:
I've retriggerd the CI. Lets see if it passes now :)

PR #729, opened 2024-02-26
fpga: enable inline config of card

n-eiling:
additionally to configuring the card in a separate block, we need to be
able to configure the card from the node config to enable libvillas
users to use the fpga node-type.


stv0g:
I've started to use the [conventional commit style](https://www.conventionalcommits.org/en/v1.0.0/) some years ago for VILLASframework. It would be nice if we can have consistent commit messages throughout the VILLAS project :)

While its not mandated by the convention, I start the commit message with an uppercase letter.

stv0g:
Can we move the fpga-related configs under `etc/fpga`?
E.g. `etc/fpga/vc707-xbar-pcie-dino/vc707-xbar-pcie-dino-v2.json`.

include/villas/node.hpp
```diff
@@ -7,6 +7,7 @@
 
 #pragma once
 
+#include <filesystem>
```
stv0g, line 10: Please note that we currently do not support the STL filesystem library due to compiler restrictions. GCC 8.3 unfortunately sometimes ships without it :(

For DPsim, I started to check this with CMake and add a standalone implementation which is compatible. But for now it would be nice to try get around without it.

lib/nodes/fpga.cpp
```diff
@@ -76,27 +86,54 @@ int FpgaNode::prepare() {
...
+    cardName = std::string(json_string_value(jsonCard));
+  } else if (json_is_object(jsonCard)) {
+    json_t *jsonCardName = json_object_get(jsonCard, "name");
+    if (jsonCardName != nullptr && !json_is_string(jsonCardName)) {
+      cardName = std::string(json_string_value(jsonCardName));
+    } else {
+      cardName = "anonymous Card";
+    }
+    auto searchPath = configPath.parent_path();
+    card = createCard(jsonCard, searchPath, vfioContainer, cardName);
```
stv0g, line 126:
```suggestion
    card = createCard(jsonCard, configPath.parent_path(), vfioContainer, cardName);
```

stv0g (changes requested):

n-eiling:
> Can we move the fpga-related configs under `etc/fpga`? E.g. `etc/fpga/vc707-xbar-pcie-dino/vc707-xbar-pcie-dino-v2.json`.

I am confused. You want it under `etc/fpga` rather than `fpga/etc`? This config file is also relevant to VILLASfpga so this would mean a duplication of the file.

include/villas/node.hpp
n-eiling: I understand. I can change all uses of std::filesystem to std::string. However, superNode and Config already use std::filesystem for a few months and I have heard no issues. GCC 8.3 is really old - I wouldn't feel too bad to drop support for this. If someone complains it's pretty easy to replace std::filesystem with std::string.

lib/nodes/fpga.cpp
n-eiling: does not work because `createCard` takes a reference. I would like to keep it this way, because in createCards we have a loop that would allocate a lot of unnecessary strings.

include/villas/node.hpp
stv0g: I already patched-out the `std::filesystem` usage in a non-public branch for about the same time :D I know GCC 8.3 is pretty old :( But its unfortunately the version which is shipped with our OPAL-RT systems. I expect they we upgrade to a newer version with the next release of the OPAL-RTLinux OS.

Can we stick with `std::string` for now? I'm also volunteering to introduce `std::filesystem` once we support it 😄 I am fully on your side, that it makes the code nicer..

lib/nodes/fpga.cpp
stv0g: Alright. Fine with me then..

stv0g:
> > Can we move the fpga-related configs under `etc/fpga`? E.g. `etc/fpga/vc707-xbar-pcie-dino/vc707-xbar-pcie-dino-v2.json`.
> 
> I am confused. You want it under `etc/fpga` rather than `fpga/etc`? This config file is also relevant to VILLASfpga so this would mean a duplication of the file.

Oh sorry. I've misinterpreted GitHubs diff.. No its perfect. Please ignore my comment.

include/villas/node.hpp
stv0g: @n-eiling I've cherry-picked the removal of `std::filesystem` in https://github.com/VILLASframework/node/pull/730

n-eiling: ok thanks. will change it on this branch.

n-eiling: ah you also changed it on this branch. even better :)

n-eiling: what about in VILLASfpga? We will never use that on OPAL targets, right?

PR #733, opened 2024-02-28
Integrate VILLAScommon submodule

stv0g:
This integrates the VILLAScommon sub-module directly into VILLASnode.
It is the first step, next we will repeat this for VILLASfpga.

/cc @windrad6 @fwege @n-eiling @al3xa23


stv0g:
Btw. I managed to keep the history of the VILLAScommon repo 🥳

stv0g:
@n-eiling

> I would say this only makes sense if we merge fpga as well, and at the same time.

I agree. Do you think we should do this in the same PR? I was planning to open another one shortly after this one is merged. But we can also do it all in one go.

> I would keep libxil as a separate repository.

I agree.

> Do I need to do anything in fpga to make this work? Running clang-format for example.

No, not necessarily. But as soon as we have integrated the VILLASfpga code into VILLASnode, I would go ahead and run clang-format over everything once again.

> What about the CI? We also need to merge that.

Yes, I already did this for VILLAScommon. Its merely a merge of the `.gitlab-ci.yml` files.

> We should probably not build fpga per default, because most setups don't need it. What do you think?

I agree as long as we enable it in the CI. Because I dont want to have any untested/compiled code in the repo.

n-eiling:
Doing it separately is fine with me. Just wanted to make sure :)

I support this PR.

The CI fails, and we need some sign-offs (Sonja and Hatim may be an issue?)

PR #739, opened 2024-03-07
node: Fix null-pointer dereference for internal loopback nodes

stv0g:
This is another approach at fixing the segmentation fault involving internal loopback nodes.

@pipeacosta This solves the underlying problem (null pointer dereference) rather than avoid calling the `villas::node::~Node()` like you did in #724.

I think this is a cleaner approach as we still call the `villas::node::Node` constructor from the `villas::node::InternalLoopbackNode` constructor.

Closes #724

### DIFF
--- a/lib/node.cpp
+++ b/lib/node.cpp
@@ -74,7 +74,10 @@ Node::~Node() {
   rtnl_cls_put(tc_classifier);
 #endif // WITH_NETEM
 
-  factory->instances.remove(this);
+  // Internal loopback nodes have no factory
+  // Only attempt removal for factories of other node-types.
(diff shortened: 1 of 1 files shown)


fwege (approved):

pipeacosta:
Great job @stv0g! Would it be possible to re-enable the `node-multiple-sources.sh` integration test?

stv0g:
Oh yes. I will push a commit to reenable them! Unfortunately, the CI is currently broken due to the CI runner beeing recondigured. Do you know more about it?

pipeacosta:
@stv0g Unfortunately, I am not yet familiar with the CI ...

PR #744, opened 2024-03-14
Update CODEOWNERS

n-eiling:
We should add more CODEOWNERS

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -13,20 +13,27 @@
 # and not the global owner(s) will be requested for a review.
 
 # FPGA
-fpga                                        @n-eiling
-/lib/nodes/fpga.cpp                         @n-eiling
-/include/villas/nodes/fpga.hpp              @n-eiling
+fpga                                                @n-eiling
+/lib/nodes/fpga.cpp                                 @n-eiling
+/include/villas/nodes/fpga.hpp                      @n-eiling
+/tools/hwdef-parse.py                               @n-eiling
+/tools/fpga-*                                       @n-eiling
+/fpga                                               @n-eiling
+/etc/examples/nodes/fpga.conf                       @n-eiling
+/etc/fpga                                           @n-eiling
+/doc/openapi/components/schemas/nodes/fpga.yaml     @n-eiling
+
 
 # PMU
-/lib/nodes/libiec61850_goose.cpp            @windrad6
-/include/villas/nodes/libiec61850_goose.hpp @windrad6
-/lib/nodes/uldaq.cpp                        @windrad6
-/include/villas/nodes/uldaq.hpp             @windrad6
-/lib/dumper.cpp                             @windrad6
-/include/villas/dumper.hpp                  @windrad6
+/lib/nodes/libiec61850_goose.cpp                    @windrad6
+/include/villas/nodes/libiec61850_goose.hpp         @windrad6
+/lib/nodes/uldaq.cpp                                @windrad6
+/include/villas/nodes/uldaq.hpp                     @windrad6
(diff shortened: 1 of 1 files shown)


stv0g:
Please dont get me wrong. But I think the codeowners file should refelct who has written / conteibuted to which areas of the code. Your change does not reflect this anymore. W.g. Felix has not conteibuted to VILLASnode at all till now.

I also believe some familiarity with the code is required for a proper review.

A review isnt just a procedural hurdle. It should improve the contribution in a meaningful way. And I havnt seen any comment or feedback in the reviews from Felix or Manuel in the past.

I know this is hard. It takes time to dig into someones elses code and understand the changes they are proposing and beeing able to procide meqningful feedback. But in my opinion, its the only way to keep a high code quality in the project.

I am happy to add more codeowners. But that requires that they also do thourough reviews and not just approve everything without actually reading the code.

n-eiling:
Ok I agree with you. It's just not practical to have you review everything. I don't think it's really necessary for you to look at new node types for example.
I actually did a review with Felix in person. And you did no review for your last PR. So we both pressed the merge without waiting for requirements box but I at least did a review (even though it might not have been as thorough as yours). So the question really is: what should the requirements be? Having a review from a current codeowner on everything obviously doesn't work.

stv0g:
How about we remove the requirement for an approval and rely on CODEOWNERS alone? I think this would unblock us as people could work alone on their area of the code. E.g. you could work on VILLASfpga without being blocked by a review. For changes which touch other parts of the code, we would still require a review as those changes potentially also affect others.

I believe for changes which are only affecting yourself, we can skip the review part. But as soon as your changes affect code of others it should be required. And those reviews should be done by those affected. E.g. Felix approving changes of code he never used is not optimal here.

I agree with you that I should not need to review all new node-types. For new code, the respective authors should add themselves directly to the CODEOWNERS. In that case, I do not need to review it. If GitHub still requires me to approve the change to the CODEOWNERS, I could do that quickly..

> I actually did a review with Felix in person. And you did no review for your last PR. 

I would be nice if you could do your review in the open, so others can participate and comment on it as well. Because for reviews done in private its really hard to reason and build trust.

> Having a review from a current codeowner on everything obviously doesn't work.

I think it works. I also try hard to do my reviews quickly. We just need more codeowners :D

>  So we both pressed the merge without waiting for requirements box

Yeah, the reviewing procedure aside. We must at least check that the pipelines pass :)

n-eiling:
> How about we remove the requirement for an approval and rely on CODEOWNERS alone?

Yes that would resolve the issue for me. Can you change the setting?

I also agree with your other points.

stv0g:
> Can you change the setting?

Yes I will do it.

Can you maybe also adjust this PR to only add yourself as the owner of the hwdef-parse script?
I think this was the only change in your last PR which required me as a reviewer.

n-eiling:
finally got around doing this.

CODEOWNERS
```diff
@@ -13,20 +13,27 @@
 # and not the global owner(s) will be requested for a review.
 
 # FPGA
-fpga                                        @n-eiling
-/lib/nodes/fpga.cpp                         @n-eiling
-/include/villas/nodes/fpga.hpp              @n-eiling
+fpga                                                @n-eiling
```
stv0g, line 16: I think this is redundant to line 21?

stv0g (changes requested):
I think we also need to ping Jonathan again about the CI runner.. :(

stv0g:
>> Can you change the setting?

> Yes I will do it.

Done :)

stv0g (approved):

PR #745, opened 2024-03-25
FPGA: Latency improvements

n-eiling:


lib/nodes/fpga.cpp
```diff
@@ -168,19 +172,93 @@ const std::string &FpgaNode::getDetails() {
...
 
 int FpgaNode::start() {
-  // enque first read
-  // dma->read(*(blockRx[0]), blockRx[0]->getSize());
+  if (getInputSignalsMaxCount() * sizeof(float) > blockRx->getSize()) {
+    logger->error("Input signals exceed block size.");
+    throw villas ::RuntimeError("Input signals exceed block size.");
+  }
+  if (lowLatencyMode) {
+    dma->readScatterGatherPrepare(*blockRx, blockRx->getSize());
```
n-eiling, line 180: @stv0g can I get your opinion on this?

Basically, when lowLatencyMode is true, I setup the read and write transactions in `start`. This means I only need to start the DMA controller in `read` and `write`. However, we cannot change the write length and do not retrieve only minimal status information from the DMA controller.

From the commit message:
This setting improves latency by remove various checks.
Use with caution! Requires read cache in FPGA design!
The common use case in VILLASfpga is that we have exactly
one write for every read and the number of exchanged signals
do not change. If this is the case, we can reuse the buffer
descriptors during reads and write, thus avoidng freeing,
reallocating and setting them up.
We set up the descriptors in start, and in write or read,
we only reset the complete bit in the buffer descriptor and
write to the tdesc register to start the DMA transfer.
Improves read/write latency by approx. 40%.

fpga/include/villas/fpga/ips/dma.hpp
```diff
@@ -128,31 +139,30 @@ class Dma : public Node {
...
+  static constexpr size_t requestedRingBdSize = 1;
   static constexpr size_t requestedRingBdSizeMemory =
       requestedRingBdSize * sizeof(XAxiDma_Bd);
-  uint32_t actualRingBdSize = XAxiDma_BdRingCntCalc(
-      XAXIDMA_BD_MINIMUM_ALIGNMENT, requestedRingBdSizeMemory);
-  std::shared_ptr<MemoryBlock> sgRingTx;
-  std::shared_ptr<MemoryBlock> sgRingRx;
+  uint32_t actualRingBdSize = 1; //XAxiDma_BdRingCntCalc(
+      //XAXIDMA_BD_MINIMUM_ALIGNMENT, requestedRingBdSizeMemory);
+  std::shared_ptr<MemoryBlock> sgRing;
```
n-eiling, line 147: These changes are so we can use a read cache in the FPGA. This avoids reading the SG descriptors from main memory (takes 1.5 us)

lib/nodes/fpga.cpp
n-eiling: Is it correct to assume that getInputSignalsMaxCount is the number of signals passed to write and read?

stv0g: For read this can be assumed as the node itself defines via its `in = { signals = (..) }` configuration which and how many signals should be read. This (currently) can not change during runtime and hence you can initialize this already in `start()`.

But I think we allow partial reads. The node might decide to provide less then the maximum number of signals per read. So this should be understood as an upper bound.

Thats also true for `write()`. If a sourcing node provide less signals than configured during its `read()`, we can have situations where the corresponding `write()` of the forewarded data sees less than the maximum number of signals.

fpga/include/villas/fpga/ips/dma.hpp
n-eiling: I removed the dead code and explained better why we use only a single BD. I added a TODO noting that we could make it configurable in the future, but right now only a 1 makes sense.

lib/nodes/fpga.cpp
n-eiling: This implementation would allow partial reads, but requires fixed length writes. If there are less signals than set during init, fastWrite would reuse the data from the previous timestep.

I documented this in a comment in the code.

fpga/lib/ips/dma.cpp
```diff
@@ -9,6 +9,9 @@
 #include <sstream>
 #include <string>
 
+#include "xilinx/xaxidma_bd.h"
```
stv0g, line 12: Please use `<>` to actually indicate that the headers should be looked up in the compilers search path rather than a relative/absolute path.

stv0g: Also sorting the includes by lib/type would be nice :)

fpga/lib/ips/dma.cpp
```diff
@@ -214,11 +213,8 @@ Dma::~Dma() {
       rxRingPtr->CyclicBd = nullptr;
     }
     // unampe SG memory Blocks
```
stv0g, line 215:
```suggestion
    // Unmap SG memory Blocks
```

fpga/lib/ips/dma.cpp
```diff
@@ -111,8 +115,15 @@ void Dma::setupScatterGatherRingRx() {
   if (ret != XST_SUCCESS)
     throw RuntimeError("Failed to clone BD template: {}", ret);
 
+  if (cyclic) {
+    /* Enable Cyclic DMA mode */
```
stv0g, line 119: Please use C++ style comments throughout the code.

For reference: https://en.cppreference.com/w/c/comment

```suggestion
    // Enable Cyclic DMA mode
```

fpga/lib/ips/dma.cpp
```diff
@@ -9,6 +9,9 @@
 #include <sstream>
 #include <string>
 
+#include "xilinx/xaxidma_bd.h"
+#include "xilinx/xaxidma_hw.h"
+#include <sys/types.h>
 #include <xilinx/xaxidma.h>
```
stv0g, line 15:
```suggestion
#include <sys/types.h>

#include <xilinx/xaxidma_bd.h>
#include <xilinx/xaxidma_hw.h>
#include <xilinx/xaxidma.h>
```

include/villas/nodes/fpga.hpp
```diff
@@ -31,15 +32,31 @@ class FpgaNode : public Node {
   std::string cardName;
   std::list<std::string> connectStrings;
 
+  // This setting improves latency by remove various checks.
```
stv0g, line 35: Thanks for the comment here :)

stv0g (changes requested):
Looks great to me. I've tried to understand all the latency tweaking you did. But thats too tough for a quick review.

I only have a few minor stylistic recommendations left.

include/villas/nodes/fpga.hpp
n-eiling: fixed grammar

fpga/lib/ips/dma.cpp
n-eiling: done

fpga/lib/ips/dma.cpp
n-eiling: done

n-eiling:
I made all the requested changes :)

stv0g (approved):

PR #780, opened 2024-05-06
FPGA: code improvements, bug fixes, etc.

n-eiling:
This is another iteration of improving the FPGA code in preparation for integration with DPsim. The current code state allows forwarding data from DPsim via VILLASnode / fpga to Dino.

For this I fixed a few errors in fpga and the DP hook, and allowed the Dino sampling rate to be configured from the fpga json.

### DIFF
--- a/etc/examples/nodes/fpga.conf
+++ b/etc/examples/nodes/fpga.conf
@@ -7,21 +7,21 @@ logging = {
 
 fpgas = {
     vc707 = {
-        interface = "pcie",
-        id = "10ee:7021",
-        slot = "0000:88:00.0",
-        do_reset = true,
-        ips = "../../fpga/etc/vc707-xbar-pcie/vc707-xbar-pcie-dino-v2.json",
-        polling =  false,
+        interface = "pcie"
+        id = "10ee:7021"
+        slot = "0000:88:00.0"
+        do_reset = true
+        ips = "../../fpga/etc/vc707-xbar-pcie/vc707-xbar-pcie-dino-v2.json"
+        polling =  false
     }
 }
 
 nodes = {
     fpga_0 = {
-        type = "fpga",
-        card = "vc707",
-        connect = ["0->3", "3->dma", "0<-dma"],
-        timestep = 10e-3,
+        type = "fpga"
+        card = "vc707"
+        connect = ["0->3", "3->dma", "0<-dma"]
+        timestep = 10e-3
     }
 }
 
--- a/etc/examples/nodes/fpga.conf
+++ b/etc/examples/nodes/fpga.conf
@@ -19,8 +19,9 @@ fpgas = {
 nodes = {
     fpga_0 = {
         type = "fpga",
-        card = "vc707"
-        connect = ["0->3", "3->dma", "0<-dma"]
+        card = "vc707",
+        connect = ["0->3", "3->dma", "0<-dma"],
+        timestep = 10e-3,
     }
 }
 
--- a/fpga/lib/ips/dino.cpp
+++ b/fpga/lib/ips/dino.cpp
@@ -140,14 +140,18 @@ void DinoAdc::configureHardware() {
 void DinoAdc::setRegisterConfig(std::shared_ptr<Register> reg,
                                 double sampleRate) {
   constexpr double dinoClk = 25e6; // Dino is clocked with 25 Mhz
+  constexpr size_t dinoRegisterTimer = 0;
+  constexpr size_t dinoRegisterScale = 1;
+  constexpr size_t dinoRegisterOffset = 2;
 
   uint32_t dinoTimerVal = static_cast<uint32_t>(dinoClk / sampleRate);
   double rateError = dinoClk / dinoTimerVal - sampleRate;
   reg->setRegister(
-      0,
+      dinoRegisterTimer,
       dinoTimerVal); // Timer value for generating ADC trigger signal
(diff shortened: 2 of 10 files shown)


lib/nodes/fpga.cpp
```diff
@@ -106,9 +119,10 @@ int FpgaNode::parse(json_t *json) {
     vfioContainer = std::make_shared<kernel::vfio::Container>();
   }
 
-  ret = json_unpack_ex(json, &err, 0, "{ s: o, s?: o, s?: b, s?: b}", "card",
-                       &jsonCard, "connect", &jsonConnectStrings,
-                       "lowLatencyMode", &lowLatencyMode);
+  ret =
+      json_unpack_ex(json, &err, 0, "{ s: o, s?: o, s?: b, s?: b, s?: f}",
+                     "card", &jsonCard, "connect", &jsonConnectStrings,
+                     "lowLatencyMode", &lowLatencyMode, "timestep", &timestep);
```
stv0g, line 125: Can we please use snake case for option names? Camel case isnt consistent with the other settings we have in VILLASnode.

lib/nodes/fpga.cpp
```diff
@@ -67,6 +69,17 @@ int FpgaNode::prepare() {
     parsedConnectString.configCrossBar(card);
   }
 
+  auto reg = std::dynamic_pointer_cast<fpga::ip::Register>(
+      card->lookupIp(fpga::Vlnv("xilinx.com:module_ref:registerif:")));
+
+  if (reg != nullptr &&
+      card->lookupIp(fpga::Vlnv("xilinx.com:module_ref:dinoif_fast:"))) {
+    // constexpr double sampleRate = 20e3; // We want to achieve a timestep of 50us
```
stv0g, line 77: Please remove dead code

etc/examples/nodes/fpga.conf
```diff
@@ -19,8 +19,9 @@ fpgas = {
 nodes = {
     fpga_0 = {
         type = "fpga",
-        card = "vc707"
-        connect = ["0->3", "3->dma", "0<-dma"]
+        card = "vc707",
+        connect = ["0->3", "3->dma", "0<-dma"],
```
stv0g, line 23: I know its unrelated to this PR, but just a recommendation for improving the config schema:

```suggestion
        connect = (
          { master = 0,     slave = 3     },
          { master = 3,     slave = "dma" },
          { master = "dma", slave = 0     }
        ),
```

`master`, `slave` could also be replaced with `from`, `to`.

This would more closer match the `paths` setting of VILLASnode itself:

```
paths = (
  {
    in = "node1"
    out = "node2"
  }
)
```

stv0g (changes requested):

include/villas/nodes/fpga.hpp
```diff
@@ -45,6 +45,15 @@ class FpgaNode : public Node {
   // Improves read/write latency by approx. 40%.
   bool lowLatencyMode;
 
+  // This sets the requested timestep in the FPGA timestep generation IP.
```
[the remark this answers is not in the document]
n-eiling: This does not generate a clock, but a pulse at a configurable interval that is (mostly) independent from the actual clock. If the FPGA is coupled with a simulation, this should be the actual time step, i.e., we generate the time step signal in FPGA and synchronize our simulator software to that signal. I think this is also how OPAL-RT does this.
Currently, its only really used for triggering conversion in the ADC though.

etc/examples/nodes/fpga.conf
n-eiling: yes makes sense. Will consider it.

fpga/lib/ips/dino.cpp
```diff
@@ -137,6 +137,25 @@ void DinoAdc::configureHardware() {
   logger->debug("ADC Ioext: Output register configured to {}", readback);
 }
 
+void DinoAdc::setRegisterConfig(std::shared_ptr<Register> reg,
+                                double sampleRate) {
+  constexpr double dinoClk = 25e6; // Dino is clocked with 25 Mhz
+
+  uint32_t dinoTimerVal = static_cast<uint32_t>(dinoClk / sampleRate);
+  double rateError = dinoClk / dinoTimerVal - sampleRate;
+  reg->setRegister(
```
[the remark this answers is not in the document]
n-eiling: The issue is, that the registers are bitstream dependent. Ideally, I would like to parse the register assignment from the XSA file. For now I switched to local constexpr to make the code more verbose.

lib/nodes/fpga.cpp
n-eiling: this I did not know. I changed it.

lib/nodes/fpga.cpp
```diff
@@ -212,7 +226,6 @@ int FpgaNode::fastWrite(Sample *smps[], unsigned cnt) {
       mem[i] = smp->data[i].i;
     }
   }
-
```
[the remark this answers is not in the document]
n-eiling: sure

include/villas/nodes/fpga.hpp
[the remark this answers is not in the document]
stv0g: Ah alright. Now I understand it. Thanks yes, in that case the naming makes more sense.

I just thought we use it in the context of Dino / ADC, where we usually refer to a sampling clock.

stv0g (approved):

PR #785, opened 2024-05-29
Fixes for NEIS paper submission

stv0g:


n-eiling (commented):
Can you summarize the changes. I think test_rtt is a pretty useful node type and the old version was basically unusable because it wasn't really documented. A lab or some hints on what setup you have been using it in would be really helpful.

PR #787, opened 2024-06-05
Various fixes for the DPsim Interface

n-eiling:
The DPsim interface required a few fixes in VILLAS.
Some are not strictly necessary, but I think the extra checks make the code more robust.

### DIFF
--- a/lib/hooks/dp.cpp
+++ b/lib/hooks/dp.cpp
@@ -191,7 +191,7 @@ public:
   virtual void prepare() {
     assert(state == State::CHECKED);
 
-    char *new_sig_name;
+    std::string new_sig_name;
 
     assert(state != State::STARTED);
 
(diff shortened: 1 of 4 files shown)


lib/hooks/dp.cpp
```diff
@@ -233,7 +233,7 @@ class DPHook : public Hook {
       signals->erase(signals->begin() + signal_index);
 
       for (int i = 0; i < fharmonics_len; i++) {
-        new_sig_name = strf("%s_harm%d", orig_sig->name, i);
+        new_sig_name = strf("%s_harm%d", orig_sig->name.c_str(), i);
```
stv0g, line 236: Would be great to switch to `fmt::format()` here

lib/nodes/fpga.cpp
```diff
@@ -341,6 +341,11 @@ int FpgaNode::slowWrite(Sample *smps[], unsigned cnt) {
 }
 
 std::vector<int> FpgaNode::getPollFDs() {
+  if (!lowLatencyMode && card && !card->polling) {
+    return card->vfioDevice->getEventfdList();
+  } else {
+    return {};
+  }
   return card->vfioDevice->getEventfdList();
```
stv0g: Seems like this is dead code? Can we remove it? Or just move the else branch out?

stv0g (changes requested):

stv0g (changes requested):

stv0g:
Thanks :)

stv0g (approved):

n-eiling:
Thanks for the fast review. I was a bit lazy on the fmt::format thing, but porting to C++ was easier than I thought.

PR #788, opened 2024-06-06
ci: improve gitlab CI

n-eiling:
The github DCO check seems to be broken. I actually liked it so maybe we can move back to it at some point. However, for now I added this DCO check to the gitlab CI: https://github.com/christophebedard/dco-check

I also added some more needs statesments to increase parallel execution of CI jobs and replaced some `CI_COMMIT_REF_NAME` with `DOCKER_TAG` (its the same value but the latter is more consistent and verbose)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,7 +40,7 @@ prepare:docker:
     matrix:
       - DISTRO: [ubuntu, debian, rocky]
       - DISTRO: fedora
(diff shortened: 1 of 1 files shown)


n-eiling:
Ah looks like now the DCO check works again.... What should we do?

stv0g:
image

Ich glaube der DCO check geht wieder.
Falls möglich würde ich den lieber direkt in GitHub belassen.

n-eiling:
ok I removed the DCO check again. I think the other changes are still useful, so let's merge them instead.

PR #790, opened 2024-06-10
bug: use float memory accesor for write to match read

IgnoreWarnings:


n-eiling:
ok I get why this is an issue for you. However, when using the Dino IP and it's digital to analog convert we need to send an integer between 0 and 0xFFFF which corrensponds to -10V to 10V. I should probably do the conversion in the FPGA but I have to think about if that really makes sense.
Anyway doing the conversion in write is probably the wrong place.

For this PR, can you also remove the code below your change and only to the double to float conversion?

IgnoreWarnings:
> ok I get why this is an issue for you. However, when using the Dino IP and it's digital to analog convert we need to send an integer between 0 and 0xFFFF which corrensponds to -10V to 10V. I should probably do the conversion in the FPGA but I have to think about if that really makes sense. Anyway doing the conversion in write is probably the wrong place.
> 
> For this PR, can you also remove the code below your change and only do the double to float conversion?

I we can also sent uint32 or any other type, but i want read and write functions to operate on the same type. Maybe change both to uint32?

n-eiling:
I changed the FPGA bitstream yesterday to convert the floats in the FPGA. This way our interface to the FPGA is always floats, which makes more sense i guess. This is also how the commercial hardware does it. Let me test this and then we can merge your changes.

PR #783, opened 2024-05-22
Support Platform Bus Fpgas

IgnoreWarnings:
Integration of platform based fpgas.

PR #791, opened 2024-06-17
Refactor: Split core make into multiple functions

IgnoreWarnings:
Seperate the core factory make method into multiple parts to improve readability, reusability and mainainability.
Verified to work as expected on miob hardware.
In addition, #783 will implement features which require the seperation of parsing and initialization.

### DIFF
--- a/fpga/lib/core.cpp
+++ b/fpga/lib/core.cpp
@@ -37,7 +37,7 @@ static std::list<Vlnv> vlnvInitializationOrder = {
     Vlnv("xilinx.com:ip:axi_iic:"),
 };
 
-std::list<IpIdentifier> CoreFactory::parseVLNV(json_t *json_ips) {
+std::list<IpIdentifier> CoreFactory::parseIpIdentifier(json_t *json_ips) {
   // Parse all IP instance names and their VLNV into list `allIps`
   std::list<IpIdentifier> allIps;
 
@@ -302,7 +302,7 @@ std::list<std::shared_ptr<Core>> CoreFactory::make(Card *card,
   }
 
   std::list<IpIdentifier> allIps =
-      parseVLNV(json_ips); // All IPs available in config
+      parseIpIdentifier(json_ips); // All IPs available in config
 
   std::list<IpIdentifier> orderedIps =
       reorderIps(allIps); // IPs ordered in initialization order
--- a/fpga/lib/core.cpp
+++ b/fpga/lib/core.cpp
@@ -54,7 +54,7 @@ std::list<IpIdentifier> CoreFactory::parseVLNV(json_t *json_ips) {
     allIps.push_back({vlnv, ipName});
   }
   return allIps;
-  }
+}
 
 std::list<IpIdentifier>
 CoreFactory::reorderIps(std::list<IpIdentifier> allIps) {
@@ -86,6 +86,7 @@ CoreFactory::reorderIps(std::list<IpIdentifier> allIps) {
 
   return orderedIps;
 }
+
 std::list<std::shared_ptr<Core>>
 CoreFactory::configureIps(std::list<IpIdentifier> orderedIps, json_t *json_ips,
                           Card *card) {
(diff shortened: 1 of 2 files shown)


IgnoreWarnings:
The e41651e commit isnt supposed to be in this PR, i will remove it tomorrow

common/include/villas/kernel/vfio_device.hpp
```diff
@@ -52,7 +52,7 @@ class Device {
   // Enable memory accesses and bus mastering for PCI device
   bool pciEnable();
 
-  int pciMsiInit(int efds[32]);
+  int pciInterruptInit(int efds[32]);
```
n-eiling, line 55: If you want to rename this then please rename all `*Msi*` functions.

fpga/lib/core.cpp
```diff
@@ -64,9 +53,14 @@ std::list<std::shared_ptr<Core>> CoreFactory::make(Card *card,
 
     allIps.push_back({vlnv, ipName});
   }
+  return allIps;
+}
 
+std::list<IpIdentifier>
+CoreFactory::reorderIps(std::list<IpIdentifier> allIps) {
```
n-eiling, line 60: does this split really make sense? Do we ever need an "unordered" list? We do not want external code to use the list returned from `parseVLNV` above, or do we?

fpga/lib/core.cpp
```diff
@@ -37,21 +37,10 @@ static std::list<Vlnv> vlnvInitializationOrder = {
...
-  std::list<IpIdentifier> allIps;     // All IPs available in config
-  std::list<IpIdentifier> orderedIps; // IPs ordered in initialization order
-
-  std::list<std::shared_ptr<Core>> configuredIps; // Successfully configured IPs
-
-  if (!card->ips.empty()) {
-    loggerStatic->error("IP list of card {} already contains IPs.", card->name);
-    throw RuntimeError("IP list of card {} already contains IPs.", card->name);
-  }
+std::list<IpIdentifier> CoreFactory::parseVLNV(json_t *json_ips) {
```
n-eiling, line 40: Maybe better names would be `parseIpIdentifier` or 'getIpIdentifier'. Why not move this to a constructor of `IpIdentfier`? Then we can here only have a `getOrderedIpList` or something like that.

fpga/lib/core.cpp
```diff
@@ -280,6 +289,28 @@ std::list<std::shared_ptr<Core>> CoreFactory::make(Card *card,
...
+  std::list<IpIdentifier> allIps =
+      parseVLNV(json_ips); // All IPs available in config
+
+  std::list<IpIdentifier> orderedIps =
+      reorderIps(allIps); // IPs ordered in initialization order
+
+  std::list<std::shared_ptr<Core>> configuredIps =
+      configureIps(orderedIps, json_ips, card); // Successfully configured IPs
+
+  (void) initIps(configuredIps, card);
```
n-eiling, line 313: you shouldn't need the `(void)`

n-eiling (changes requested):

n-eiling:
please fix DCOs and I would also like a review from @stv0g. I'm not sure if the logger use is correct.

fpga/lib/core.cpp
IgnoreWarnings: Making the ordering of ips optional is beneficial, because later refactors might be able to completely remove ordered initialization dependencies.

fpga/lib/core.cpp
IgnoreWarnings: I renamed the method to parseIpIdentifier. I dislike "get", since it doesnt get any member.
Core constructor should not parse anything, since it would break the single responsiblity principle.

fpga/lib/core.cpp
IgnoreWarnings: Yes. I removed it.

fpga/lib/core.cpp
n-eiling: I meant the IpIdentifier Constructor. But maybe we don't want to bring the jansson dependency into that file.

n-eiling (approved):

PR #793, opened 2024-06-28
FPGA: move float conversion to fpga and various fixes

n-eiling:


fpga/lib/ips/dino.cpp
```diff
@@ -141,23 +141,40 @@ void DinoAdc::setRegisterConfig(std::shared_ptr<Register> reg,
...
-                   -0.001615254F);                // Scale factor for ADC value
-  reg->setRegister(dinoRegisterOffset, 10.8061F); // Offset for ADC value
-  uint32_t rate = reg->getRegister(0);
-  float scale = reg->getRegisterFloat(1);
-  float offset = reg->getRegisterFloat(2);
-  logging.get("Dino")->info("Check: Register configuration: Rate: {}, Scale: "
-                            "{}, Offset: {}, Rate-Error: {} Hz",
-                            rate, scale, offset, rateError);
+  reg->setRegister(dinoRegisterAdcScale,
+                   -0.001615254F); // Scale factor for ADC value
```
stv0g, line 157: Is there an equation on how this was calculated. Magic constants are not a nice code style..

fpga/lib/ips/dino.cpp
```diff
@@ -141,23 +141,40 @@ void DinoAdc::setRegisterConfig(std::shared_ptr<Register> reg,
...
+                   3448.53852516F); // Scale factor for DAC value
+  reg->setRegister(dinoRegisterDacOffset, 32767.5F); // Offset for DAC value
+  uint32_t rate = reg->getRegister(dinoRegisterTimer);
+  float adcScale = reg->getRegisterFloat(dinoRegisterAdcScale);
+  float adcOffset = reg->getRegisterFloat(dinoRegisterAdcOffset);
+  float dacScale = reg->getRegisterFloat(dinoRegisterDacScale);
+  float dacOffset = reg->getRegisterFloat(dinoRegisterDacOffset);
+  uint32_t dacExternalTrig = reg->getRegister(dinoRegisterDacExternalTrig);
+  uint32_t stsActive = reg->getRegister(dinoRegisterStsActive);
+  logging.get("Dino")->info(
```
stv0g, line 172: Can we have a `logger` member to avoid the `logging.get` calls?

stv0g (changes requested):

stv0g:
Looks good to me with only two smaller notes/nits

fpga/lib/ips/dino.cpp
n-eiling: Why do you prefer this? This is a single logging call and it's only during setup - so performance doesn't matter. And this is also less code than having the member.
Also I still don't like all these logger objects in general :P

stv0g: Okay if it's just a single call then it's okay I guess... For more or ones spread across multiple functions I would move it to a member or a local variable in case they are in the same function...

fpga/lib/ips/dino.cpp
stv0g: Would it be okay to move it to a configuration setting now? Maybe with your magic constant as a default value? I think it's crucial for future users to understand the meaning and how it was determined. So a sentence or two of documentation would not hurt here..

n-eiling:
@stv0g can you approve this? :)

n-eiling:
@stv0g :wave:

stv0g:
Hi @n-eiling

There is still some dead-code and my comment about the hard-coded calibration values which was not addressed yet.

n-eiling:
well, I can remove the dead code. But I think there are already sufficient comments for the calibration values. This should all change in the future anyways. I want to move these values to the DINO EEPROM.

n-eiling:
I this better? :)

fpga/lib/ips/dino.cpp
n-eiling: I added a more detailed comment.

stv0g (approved):
Yes, thanks :)

PR #800, opened 2024-07-29
Fix static initialization

### LINKED ISSUE #799: static logger initialization and subsequent static initializations using logging.get is undefined behavior
We initialize `villas::logging` as a static member in `log.cpp`. At the same time several other objects are created as static objects. Some of them (e.g., `MemoryManager`, `DirectedGraph`, `HostRamAllocator`) use the static object to create their own logging instances, i.e., `villas::logging.get()`.
The initialization order of these objects is undefined and thus what we are doing is undefined behavior.
If I activate link time optimization (`-lto`) I get segmentation faults because the initialization order is different.

I would suggest to move the initialization of these objects to an explicit initialization function, where we can make the initialization order explicit and call this function for example from a library constructor or via a single static initialization.

I only researched this briefly so I may have misunderstood some of this.

n-eiling:
fixes #799 

I replaced the static constructor call of `villas::logging` and `HostRamAllocator::allocator` to a function that does lazy initialization (i.e., on the first call).
This avoid the undefined behavior mentioned in the issue and even avoid an explicit initialization order.
The implementation leak heap memory that will be implicitly cleaned up when the application finished. While currently not a problem, the destructors of Log and HostRamAllocator should not assume the existence of other static objects, because the deinitialization order is still undefined.

I implemented the C++ Foundaitions recommendation for this issue (see https://isocpp.org/wiki/faq/ctors#static-init-order)

This PR allows using link time optimization (`-lto` gcc flag) with VILLASnode.

### DIFF
--- a/common/include/villas/memory.hpp
+++ b/common/include/villas/memory.hpp
@@ -116,7 +116,7 @@ public:
     // CRTP
     derivedAlloc = static_cast<DerivedAllocator *>(this);
     std::string loggerName = fmt::format("memory:", derivedAlloc->getName());
-    logger = logging.get(loggerName);
+    logger = Log::get(loggerName);
 
     // Default deallocation callback
     free = [&](MemoryBlock *mem) {
@@ -249,10 +249,12 @@ public:
     allocateBlock(size_t size);
   };
 
-  static HostRamAllocator &getAllocator() { return allocator; }
-
-private:
-  static HostRamAllocator allocator;
+  static HostRamAllocator &getAllocator() {
+    // This will "leak" memory. Because we don't have a complex destructor it's okay
+    // that this will be cleaned up implicitly on program exit.
+    static auto allocator = new HostRamAllocator();
+    return *allocator;
+  }
 };
 
 class HostDmaRam {
--- a/common/include/villas/graph/directed.hpp
+++ b/common/include/villas/graph/directed.hpp
@@ -31,7 +31,7 @@ public:
   using Path = std::list<EdgeIdentifier>;
 
   DirectedGraph(const std::string &name = "DirectedGraph")
-      : lastVertexId(0), lastEdgeId(0), logger(logging.get(name)) {}
+      : lastVertexId(0), lastEdgeId(0), logger(Log::get(name)) {}
 
   std::shared_ptr<VertexType> getVertex(VertexIdentifier vertexId) const {
     // Cannot use [] operator, because creates non-existing elements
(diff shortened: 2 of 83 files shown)


common/include/villas/memory.hpp
```diff
@@ -249,10 +249,12 @@ class HostRam {
     allocateBlock(size_t size);
   };
 
-  static HostRamAllocator &getAllocator() { return allocator; }
-
-private:
-  static HostRamAllocator allocator;
+  static HostRamAllocator &getAllocator() {
```
stv0g, line 252: Could we use the same approach for the logging class as well?

E.g. have a static function, with a static private/projected member as singleton?

```cpp
Log::getLogger("some_name")
```

is nicer than:

```cpp
logging().get("some_name")
```

I also prefer to have an harmonized way to create those singletons.

stv0g (changes requested):
I only have a small request :)

But thanks a lot for the effort and debugging the issue :)

common/include/villas/memory.hpp
n-eiling: ok. I wanted to be minimally invasive, but this sounds sensible.

n-eiling:
I updated the PR. Was this what you thought of?

stv0g:
Yes great :)

Just wondering did you want to keep the inconsistency intentionally?

`HostRamAllocator::getAllocator` vs `Log::get()`

I would opt for one of the following:

- `HostRamAllocator::get()` & `Log::get()`
- `HostRamAllocator::getAllocator()` & `Log::getLogger()`

n-eiling:
Hmm difficult ... The choice is
`HostRam::HostRamAllocator::getInstance()`
or
`HostRam::HostRamAllocator::get()`
or
`HostRam::getAllocator()`

I think I still prefer the last one.
For Log there is now
`Log::getInstance()` to get what was originally `villas::logger`, and `Log::get()` for what was originally `villas::logger.get()`. This avoids one function call compared to `Log::getInstance().get()`, which should be slightly better for performance.

windrad6 (approved):

stv0g:
> I think I still prefer the last one.

Okay, then lets go also with `Log::getLogger()`. Especially if there is also `Log::getInstance()`. Because otherwise, I would ask myself: what do we `get()` here?

n-eiling:
Was this ok now, or do you want me to change anything?

n-eiling:
@stv0g

stv0g (approved):

PR #803, opened 2024-07-30
BugFix: Make parsing of baseaddress optional

IgnoreWarnings:
The parameters should not be expected to contain a baseaddress. I forgot to add this in the original PR.


n-eiling:
can you remove the debug output? I think this is not necessary.

stv0g:
Please use the 'a?: I' format string in json_unpack to indicate optional fields. Take a look at the remaining VILLASnode code. We use quite often for optional fields.

PR #808, opened 2024-08-08
FPGA: Device and Driver

IgnoreWarnings:
This PR adds devices and drivers with features of binding, unbinding and probing. The IpDevice class adds additional features for devices which correspond to an ip on the fpga, such as getting the ip name and baseaddress.

Used by https://github.com/VILLASframework/node/pull/783.

### DIFF
--- a/common/include/villas/kernel/devices/device.hpp
+++ b/common/include/villas/kernel/devices/device.hpp
@@ -1,0 +1,33 @@
+/* Interface for Linux/Unix devices.
+ *
+ * Author: Pascal Bauer <pascal.bauer@rwth-aachen.de>
+ *
+ * SPDX-FileCopyrightText: 2023-2024 Pascal Bauer <pascal.bauer@rwth-aachen.de>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <filesystem>
+#include <optional>
+#include <villas/kernel/devices/driver.hpp>
+
+namespace villas {
+namespace kernel {
+namespace devices {
+
+class Device {
+public:
+  virtual ~Device(){};
+
+  virtual std::optional<std::unique_ptr<Driver>> driver() const = 0;
+  virtual std::optional<int> iommu_group() const = 0;
+  virtual std::string name() const = 0;
+  virtual std::filesystem::path override_path() const = 0;
+  virtual std::filesystem::path path() const = 0;
+  virtual void probe() const = 0;
+};
+
+} // namespace devices
+} // namespace kernel
+} // namespace villas
--- a/common/include/villas/kernel/devices/driver.hpp
+++ b/common/include/villas/kernel/devices/driver.hpp
@@ -1,0 +1,29 @@
+/* Interface for device drivers. OS/platform independend. 
+ * Implemented for Linux/Unix drivers in linux_driver.hpp
+ *
+ * Author: Pascal Bauer <pascal.bauer@rwth-aachen.de>
+ *
+ * SPDX-FileCopyrightText: 2023-2024 Pascal Bauer <pascal.bauer@rwth-aachen.de>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+namespace villas {
+namespace kernel {
+namespace devices {
+
+class Device;
+
+class Driver {
+public:
+  virtual void attach(const Device &device) const = 0;
+  virtual void bind(const Device &device) const = 0;
+  virtual std::string name() const = 0;
+  virtual void override(const Device &device) const = 0;
+  virtual void unbind(const Device &device) const = 0;
+};
+
+} // namespace devices
+} // namespace kernel
+} // namespace villas
--- a/common/include/villas/kernel/devices/ip_device.hpp
+++ b/common/include/villas/kernel/devices/ip_device.hpp
@@ -22,7 +22,8 @@ public:
 
 private:
   IpDevice() = delete;
-  IpDevice(const std::filesystem::path valid_path) //! Dont allow unvalidated paths
+  IpDevice(
+      const std::filesystem::path valid_path) //! Dont allow unvalidated paths
       : PlatformDevice(valid_path){};
 
 public:
--- a/common/include/villas/kernel/devices/ip_device.hpp
+++ b/common/include/villas/kernel/devices/ip_device.hpp
@@ -1,0 +1,35 @@
+/* Linux/Unix device which represents an IP component of a FPGA.
+ *
+ * Author: Pascal Bauer <pascal.bauer@rwth-aachen.de>
+ *
+ * SPDX-FileCopyrightText: 2023-2024 Pascal Bauer <pascal.bauer@rwth-aachen.de>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <filesystem>
+#include <villas/kernel/devices/platform_device.hpp>
+
+namespace villas {
+namespace kernel {
+namespace devices {
+
+class IpDevice : public PlatformDevice {
+public:
+  static IpDevice from(const std::filesystem::path unsafe_path);
+  static bool is_path_valid(const std::filesystem::path unsafe_path);
+
+private:
+  IpDevice() = delete;
+  IpDevice(const std::filesystem::path valid_path) //! Dont allow unvalidated paths
+      : PlatformDevice(valid_path){};
+
+public:
+  size_t addr() const;
+  std::string ip_name() const;
+};
+
+} // namespace devices
+} // namespace kernel
+} // namespace villas
--- a/common/include/villas/utils.hpp
+++ b/common/include/villas/utils.hpp
@@ -8,10 +8,10 @@
 
 #pragma once
 
+#include <filesystem>
 #include <list>
 #include <string>
 #include <vector>
-#include <filesystem>
 
 #include <cassert>
 #include <cstdint>
@@ -53,7 +53,7 @@
 #ifdef ALIGN
 #undef ALIGN
 #endif
-#define ALIGN(x, a) ALIGN_MASK(x, (uintptr_t)(a) - 1)
+#define ALIGN(x, a) ALIGN_MASK(x, (uintptr_t)(a)-1)
 #define ALIGN_MASK(x, m) (((uintptr_t)(x) + (m)) & ~(m))
 #define IS_ALIGNED(x, a) (ALIGN(x, a) == (uintptr_t)x)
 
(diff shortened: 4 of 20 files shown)


n-eiling (changes requested):
I think this should go into `common/lib/kernel`. Also there we already have a class `Device` in `pci.hpp`. Maybe it makes sense to make a top class `Device` from which you derive `PciDevice` and `PlatformDevice` (instead of `IpDevice`)?

Doesn't this make some of the functions in `common/kernel` obsolete? Can we remove something there?
Also please add me to the `CODEOWNERS` file for the new files you created.

common/include/villas/kernel/devices/ip_device.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <filesystem>
+#include <villas/kernel/devices/platform_device.hpp>
+
+namespace villas {
+namespace kernel {
+namespace devices {
+
+class IpDevice : public PlatformDevice {
+public:
```
n-eiling, line 19: remove one `public:`

common/include/villas/kernel/devices/generic_driver.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <filesystem>
+#include <fstream>
+#include <iostream>
+#include <villas/kernel/devices/driver.hpp>
+
+namespace villas {
+namespace kernel {
+namespace devices {
+
+class GenericDriver : public Driver {
```
n-eiling, line 20: Why don't you implement the `GenericDriver` in `Driver`?

common/include/villas/kernel/devices/utils.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <filesystem>
+#include <string>
+#include <vector>
+
+namespace villas {
+namespace kernel {
+namespace devices {
+namespace utils {
+
+void write_to_file(std::string data, const std::filesystem::path file);
```
n-eiling, line 20: doesn't it make sense to have this somewhere more general? @stv0g maybe you know a good place? `common/include/villas/utils` maybe?

common/lib/kernel/devices/generic_driver.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+using villas::kernel::devices::Device, villas::kernel::devices::GenericDriver;
+using villas::kernel::devices::utils::write_to_file;
+
+void GenericDriver::attach(const Device &device) const {
+  if (device.driver().has_value()) {
+    device.driver().value()->unbind(device);
+  }
+  this->override(device);
+  device.probe();
```
n-eiling, line 22: I think the writes to the sys files can fail with a few different errors that contain information on what happened. Maybe we can put this information in an exception.

common/lib/kernel/devices/ip_device.cpp
```diff
@@ -0,0 +1,55 @@
...
+  if (!is_path_valid(unsafe_path))
+    throw std::runtime_error(
+        "Path \"" + unsafe_path.u8string() +
+        "\" failed validation as IpDevicePath \"[adress in hex].[name]\". ");
+  return IpDevice(unsafe_path);
+}
+
+std::string IpDevice::ip_name() const {
+  int pos = name().find('.');
+  return name().substr(pos + 1);
```
n-eiling, line 27: shouldn't you check if the string is pos+1 long?

common/lib/kernel/devices/ip_device.cpp
```diff
@@ -0,0 +1,55 @@
...
+  size_t addr = 0;
+  ss >> addr;
+
+  return addr;
+}
+
+bool IpDevice::is_path_valid(const std::filesystem::path unsafe_path) {
+  // Split the string at last slash
+  int pos = unsafe_path.u8string().rfind('/');
+  std::string assumed_device_name = unsafe_path.u8string().substr(pos + 1);
```
n-eiling, line 46: there is also `std::filesystem::path::filename()`. Why not use that?

common/lib/kernel/devices/platform_device.cpp
```diff
@@ -0,0 +1,54 @@
...
+  write_to_file(this->name(), this->m_probe_path);
+}
+
+std::filesystem::path PlatformDevice::override_path() const {
+  return this->m_override_path;
+}
+
+std::string PlatformDevice::name() const {
+  size_t pos = this->m_path.u8string().rfind('/');
+  return this->m_path.u8string().substr(pos + 1);
```
n-eiling, line 53: `.filename()`

common/lib/kernel/devices/utils.cpp
```diff
@@ -0,0 +1,48 @@
...
+    // Write to file
+    outputFile << data;
+    outputFile.close();
+  } else {
+    throw std::filesystem::filesystem_error("Cannot open outputfile",
+                                            std::error_code());
+  }
+}
+
+std::vector<std::string>
```
n-eiling, line 32: why not a reference?

n-eiling (changes requested):

common/include/villas/kernel/devices/utils.hpp
stv0g: I agree. We already have https://github.com/VILLASframework/node/blob/master/common/lib/utils.cpp

common/lib/kernel/devices/ip_device.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <regex>
+#include <stdexcept>
+
+#include <villas/kernel/devices/ip_device.hpp>
+
+using villas::kernel::devices::IpDevice;
+
+IpDevice IpDevice::from(const std::filesystem::path unsafe_path) {
+  if (!is_path_valid(unsafe_path))
+    throw std::runtime_error(
```
stv0g, line 19: I think we already have our own Exception classes which use `fmt::format` to format the exception message.

stv0g (changes requested):

n-eiling (approved):

common/lib/kernel/devices/generic_driver.cpp
```diff
@@ -0,0 +1,40 @@
+/* GenericDriver
```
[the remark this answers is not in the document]
stv0g: This hasnt been addressed?

common/include/villas/kernel/devices/device.hpp
```diff
@@ -0,0 +1,33 @@
+/* Interface for Linux/Unix devices.
+ *
+ * Author: Pascal Bauer <pascal.bauer@rwth-aachen.de>
+ *
+ * SPDX-FileCopyrightText: 2023-2024 Pascal Bauer <pascal.bauer@rwth-aachen.de>
```
stv0g, line 5: Just wondering: has this work been done as part of a contract with RWTH? Shouldnt we add RWTH as the copyright holder in this case?

common/include/villas/kernel/devices/driver.hpp
```diff
@@ -0,0 +1,28 @@
+/* Interface for Linux/Unix device drivers
```
stv0g, line 1: Please end comments with a period `.`

common/include/villas/kernel/devices/generic_driver.hpp
```diff
@@ -0,0 +1,52 @@
+/* GenericDriver - Works for standard Linux drivers
```
stv0g, line 1:
```suggestion
/* GenericDriver - Works for standard Linux drivers.
```

common/include/villas/kernel/devices/driver.hpp
stv0g, line 1:
```suggestion
/* Interface for Linux/Unix device drivers.
```

common/include/villas/kernel/devices/generic_driver.hpp
stv0g: This is not descriptive at all. Please briefly summarize what a generic driver is.

Does it abstract a Linux Kernel device? FPGA Internal IP component? VFIO abstraction?

common/include/villas/kernel/devices/ip_device.hpp
```diff
@@ -0,0 +1,35 @@
+/* IpDevice: Linux/Unix device which represents an IP component of a fpga
```
stv0g, line 1:
```suggestion
/* IpDevice: Linux/Unix device which represents an IP component of a FPGA.
```

common/lib/utils.cpp
```diff
@@ -351,5 +355,35 @@ bool isPrivileged() {
...
+                                            std::error_code());
+  }
+}
+
+std::vector<std::string> read_names_in_directory(const std::string &name) {
+  DIR *directory = opendir(name.c_str());
+
+  struct dirent *dp;
+  std::vector<std::string> names;
+  dp = readdir(directory);
```
stv0g, line 376: Please be consistent in the APIs you are using.

I think instead of libc, you should use [`std::filesystem::directory_iterator`](https://en.cppreference.com/w/cpp/filesystem/directory_iterator) here..

common/lib/kernel/devices/platform_device.cpp
```diff
@@ -0,0 +1,49 @@
+/* Platform Device
```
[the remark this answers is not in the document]
stv0g: This has not been addressed...

common/lib/kernel/devices/ip_device.cpp
```diff
@@ -0,0 +1,53 @@
+/* IpDevice
```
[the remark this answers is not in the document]
stv0g: This has not been addressed...

common/include/villas/utils.hpp
```diff
@@ -211,6 +212,9 @@ template <class... Ts> struct overloaded : Ts... {
 // explicit deduction guide (not needed as of C++20)
```
stv0g, line 212:
```suggestion
// Explicit deduction guide (not needed as of C++20)
```

stv0g (changes requested):
A little request: please do not resolve conversations about points raised by a reviewer yourself. Its the task of the reviewer to check if the points have been addressed.

common/include/villas/kernel/devices/device.hpp
n-eiling: I felt like the result of our discussion on this topic was that we do not enforce whether we list a company or a person here. I'm still of the opinion that the university cannot be the copyright owner.

stv0g: Okay, I am giving up :D

stv0g: Little background on why I am so persistent about this topic:

We had several cases in the past in which software has been relicensed.
Doing so requires the consent from all contributors (or their companies which hold the rights on their contributions). With contributions like the one here, its impossible to distinguish whether it has been made by an individual or somebody employed.

So in case of any copyright related questions we will always have to reach our to the persond directly. And you know that this can be difficult at times for students who have long left the institute.

stv0g:
> Lets only comment the headers to avoid duplication.

This will break consistency. If its decided to add descriptions/comments only to headers we need to do this in all source files of the repo. As this would be a larger change, we will need to do it in a separete MR. So for now, please keep consistent with the rest of the code base. Thanks :)

stv0g (approved):

PR #809, opened 2024-08-08
FPGA: Add new jsons

IgnoreWarnings:
Add dino-aurora and smmu jsons


n-eiling (changes requested):
Two comments: I would remove the date from the aurora_dino json to stay consistent with the other json. And didn't we remove the dino IPs from the aurora_dino json? If this is still removed I would not add this until it is fully supported.

n-eiling:
We removed something from the json so that the Dino class is not created. So technically the `dino` in the name is not correct.

PR #567, opened 2022-12-02
Real-time optimizations

stv0g:
In GitLab by @stv0g on Mar 15, 2022, 17:33

Closes #330

PR #811, opened 2024-08-09
Rt optimizations rebased

n-eiling:
rebased variant of #567


stv0g:
@n-eiling Do you mind giving this a review? I am curious about your opinion on these changes.

n-eiling:
Regarding the timer changes: It looks like it changes many calls to Task to use the default parameter, which currently is `Task(int clock = CLOCK_REALTIME);`. I would assume we want CLOCK_MONOTONIC though. CLOCK_REALTIME can change during runtime, e.g., because NTP/chronyd updates the time - CLOCK_MONOTONIC will not.
Regarding the prefaults: I didn't test them - I assumed you researched how to do this correctly before creating the PR?
Otherwise the PR adds a bit of error handling.

stv0g:
> I would assume we want CLOCK_MONOTONIC though.

I agree. Dunno, why I changed this back then.. Maybe I change the default to the monotonic clock as this is usually preferred.

n-eiling:
ok let me change this to a draft and I'll try to test / improve when I get the time.

stv0g (approved):

PR #813, opened 2024-08-13
fix(iec61850-9-2): Fix Sampled Values node-type for NEIS paper

stv0g:


lib/nodes/iec61850_sv.cpp
```diff
@@ -320,6 +315,7 @@ int villas::node::iec61850_sv_start(NodeCompat *n) {
 
     i->out.publisher =
         SVPublisher_createEx(&comm_params, i->interface, i->out.vlan.enabled);
+    // SVPublisher_create(nullptr, i->interface);
```
n-eiling, line 318: dead code!

stv0g: :D Sorry, I will fix this.

PR #820, opened 2024-09-20
VFIO: Add method getNumberIrqs()

IgnoreWarnings:
Add a method to get the number of interrupts on a vfio device.


n-eiling (approved):

stv0g:
Hey sorry for beeing picky but I find "Number" to be confusing in this context. I think using "Count" would be more meaningful and would avoid confusion with IRQ numbers...

Can we change this?

PR #822, opened 2024-10-16
node-ethercat: Fix #include after opened namespace

PJungkamp:
The `ethercat_config.hpp` header was included from `ethercat.hpp` after opening the `villas::node` namespace. This made all definitions originating from that header including the transitive definitions from ethercat's `ecrt.h` part of the `villas::node` namespace.

The especially problematic part was that `ecrt.h` itself has an `#include <netinet/in.h>` introduced in [1]. This caused the libc definitions for `sa_familt_t`, `sockaddr_in` and `sockaddr_in6` to be moved into the `villas::node` namespace, causing build error on recent `ethercat` versions.

[1]: https://gitlab.com/etherlab.org/ethercat/-/commit/c2f9baf96ae0c0619dead65e19d3ace0e66f031b

### DIFF
--- a/include/villas/nodes/ethercat.hpp
+++ b/include/villas/nodes/ethercat.hpp
@@ -18,21 +18,16 @@
 #include <villas/queue_signalled.h>
 #include <villas/task.hpp>
 
-namespace villas {
-namespace node {
+// Include hard-coded Ethercat Bus configuration
+#include <villas/nodes/ethercat_config.hpp>
+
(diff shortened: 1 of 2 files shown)


PJungkamp:
I always screw up the DCO..... @stv0g Let's wait for the CI and get this in ASAP.

stv0g (approved):
Thanks a lot for the fix :)

Such a stupid mistake..

PR #823, opened 2024-10-22
Pi5 runner

al3xa23:
Hi, 

just realized that the old PR from this branch was closed. I rebased this branch so that we can merge it.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,7 +4,7 @@
 variables:
   GIT_SUBMODULE_STRATEGY: recursive
   GIT_DEPTH: 0
-  DISTRO: fedora # standard distro
+  DISTRO: fedora # Standard distribution
   DOCKER_FILE: packaging/docker/Dockerfile.${DISTRO}
   DOCKER_TAG: ${CI_COMMIT_REF_NAME}
   DOCKER_IMAGE: registry.git.rwth-aachen.de/acs/public/villas/node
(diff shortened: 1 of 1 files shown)


stv0g (changes requested):
We should include the Arm64 Docker image into the manifest here:
https://github.com/VILLASframework/node/blob/248bbf55c7228c1ab106e1efdc178d56cdc3a157/.gitlab-ci.yml#L216

It would also nice if we could stick with the convention and use lowercase tags to select the runners. `PI5` falls outside the convention here..

al3xa23:
arm64 is inclued. But for some reason, the tag PI5 only works in capital letters.

stv0g:
@al3xa23 The Runner configuration in GitLab was missing the correct (lowercase) tag. I have fixed this:

.gitlab-ci.yml
```diff
@@ -184,6 +184,12 @@ pkg:docker:
         PLATFORM: linux/amd64
         ARCH: x86_64
         TRIPLET: x86_64-linux-gnu
+        TAG: docker
+      - DISTRO: debian
+        PLATFORM: linux/arm64/v8
+        ARCH: arm64
+        TRIPLET: aarch64-linux-gnu
+        TAG: PI5
```
stv0g, line 192:
```suggestion
        TAG: pi5
```

stv0g (approved):

PR #826, opened 2024-10-27
Improving FPGA communication

n-eiling:
- Various fixes for FPGA use cases.
- New IP driver axis_cache


fpga/lib/ips/axis_cache.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  for (size_t i = 1; i < registerNum; i++) {
+    buf = getRegister(i);
+    if (buf != 0x00FF00FF) {
+      logger->error("Register {}: 0x{:08x} != 0x{:08x}", i, buf, i);
+      return false;
+    }
+  }
+
+  // Reset Registers
```
stv0g, line 43: That comment can be removed

fpga/lib/ips/axis_cache.cpp
```diff
@@ -0,0 +1,87 @@
...
+  invalidate();
+  return true;
+}
+
+bool AxisCache::check() {
+
+  logger->debug("Checking register interface: Base address: 0x{:08x}",
+                getBaseAddr(registerMemory));
+  uint32_t buf;
+  // we shouldn't change the rate register, because this can lead to hardware fault, so start at 1
```
stv0g, line 30:
```suggestion

  // We should not change the rate register, because this can lead to hardware fault, so start at 1
```

fpga/lib/ips/axis_cache.cpp
```diff
@@ -0,0 +1,87 @@
+/* Driver for AXI Stream read cache. This module is used to lower latency of
```
stv0g, line 1:
```suggestion
/* Driver for AXI Stream read cache.
 *
 * This module is used to lower latency of
```

fpga/lib/ips/axis_cache.cpp
```diff
@@ -0,0 +1,87 @@
...
+}
+
+void AxisCache::invalidate() {
+  setRegister(0, 1U << 31);
+  logger->info("invalidated AXIS cache.");
+}
+
+void AxisCache::setRegister(size_t reg, uint32_t value) {
+  if (reg >= registerNum) {
+    logger->error("Register index out of range: {}/{}", reg, registerNum);
```
stv0g, line 60: I think this pattern is a bit strange.. Can we pass the formatted error message to the exception instead?

I think logging should happen while catching the exception.

(Also applies to all other expcetions thrown in this class)

fpga/lib/ips/dino.cpp
```diff
@@ -140,43 +140,63 @@ void DinoAdc::configureHardware() {
...
-  reg->setRegister(
-      dinoRegisterTimer,
-      dinoTimerVal); // Timer value for generating ADC trigger signal
   // The following are calibration values for the ADC and DAC. Scale
   // sets an factor to be multiplied with the input value. This is the
   // raw 16 bit ADC value for the ADC and the float value from VILLAS for
   // the DAC. Offset is a value to be added to the result of the multiplication.
   // All values are IEE 754 single precision floating point values.
+  // Calibration for ADC filter with C=330pF and R=2,2kOhm
+  // FIXME: These values should be read from the FPGA or configured via the configuration file.
```
stv0g, line 169:
```suggestion
  // TODO: These values should be read from the FPGA or configured via the configuration file.
```

lib/hooks/cast.cpp
```diff
@@ -93,7 +93,8 @@ class CastHook : public MultiSignalHook {
 static char n[] = "cast";
 static char d[] = "Cast signals types";
 static HookPlugin<CastHook, n, d,
-                  (int)Hook::Flags::NODE_READ | (int)Hook::Flags::PATH>
+                  (int)Hook::Flags::NODE_READ | (int)Hook::Flags::NODE_WRITE |
+                      (int)Hook::Flags::PATH>
```
stv0g, line 97: Is this indentation really correct?

fpga/lib/ips/dino.cpp
```diff
@@ -140,43 +140,63 @@ void DinoAdc::configureHardware() {
...
   constexpr size_t dinoRegisterStsActive = 5;
   constexpr size_t dinoRegisterDacScale = 6;
   constexpr size_t dinoRegisterDacOffset = 7;
+  constexpr size_t dinoRegisterTimerPreThresh = 8;
+
+  // -1 because the timer counts from 0 to the value set in the register. Should really be fixed in hardware.
+  uint32_t dinoTimerVal = static_cast<uint32_t>(dinoClk / sampleRate) - 1;
+  uint32_t dinoDacDelayCycles = static_cast<uint32_t>(dinoClk * dinoDacDelay);
+  double rateError = dinoClk / (dinoTimerVal + 1) - sampleRate;
+  // Timer value for generating ADC trigger signal
```
stv0g, line 160:
```suggestion

  // Timer value for generating ADC trigger signal
```

fpga/lib/ips/dino.cpp
```diff
@@ -140,43 +140,63 @@ void DinoAdc::configureHardware() {
...
+                   32772.159015058445F); // Offset for DAC value
+  reg->setRegister(dinoRegisterDacExternalTrig,
+                   (uint32_t)0x0); // External trigger for DAC
+  if (dinoTimerVal > dinoDacDelayCycles) {
+    reg->setRegister(dinoRegisterTimerPreThresh,
+                     dinoTimerVal - dinoDacDelayCycles);
+  } else {
+    reg->setRegister(dinoRegisterTimerPreThresh, dinoTimerVal);
+  }
   uint32_t rate = reg->getRegister(dinoRegisterTimer);
```
stv0g, line 186:
```suggestion

  uint32_t rate = reg->getRegister(dinoRegisterTimer);
```

fpga/lib/ips/dino.cpp
```diff
@@ -140,43 +140,63 @@ void DinoAdc::configureHardware() {
...
+                   -11.365293957141239F); // Offset for ADC value
   reg->setRegister(dinoRegisterDacScale,
-                   3448.53852516F); // Scale factor for DAC value
-  reg->setRegister(dinoRegisterDacOffset, 32767.5F); // Offset for DAC value
+                   3204.7355379027363F); // Scale factor for DAC value
+  reg->setRegister(dinoRegisterDacOffset,
+                   32772.159015058445F); // Offset for DAC value
+  reg->setRegister(dinoRegisterDacExternalTrig,
+                   (uint32_t)0x0); // External trigger for DAC
+  if (dinoTimerVal > dinoDacDelayCycles) {
```
stv0g, line 180:
```suggestion

  if (dinoTimerVal > dinoDacDelayCycles) {
```

stv0g (dismissed):

lib/hooks/cast.cpp
n-eiling: This is what clang-format enforces. I don't really get why. I'll separate the flags to make it look nicer.

fpga/lib/ips/axis_cache.cpp
n-eiling: OK so this is again a topic for our contributor guidelines. I started adding these logging outputs because catching the exception at a higher level makes it very difficult to understand where an error occurred (compounded by the fact that we don't use static loggers and therefore cannot output line and file information).
In the case here, if I remove the logger output, I would only get the error after all the destructors have been called. And because we do a lot of logging in destructors it's really difficult to undestand what happens due to the error and what is normal behavior.

It may be my C background, but I like to handle errors where they occur, because you lose information if you handle it somewhere else.
Would it be better if I throw an exception without message to avoid the duplication?

stv0g: > Would it be better if I throw an exception without message to avoid the duplication?

Mhh, but then we loose the context of the exception :/

I would rather say, we should not log the error as this also removes the chance for an exception handler to silently ignore or handle the error as it always yields an error message in the log.

Can we pass more context to the exception? E.g. using the existing `Exception` class which VILLASnode provides, and extending it with an additional constructor which receives the `__FILE__`, `__LINE__` macros?

stv0g:
Thanks for addressing the nits about the code-style. Unfortunately, I think the DCO check isnt happy anymore about it..

fpga/lib/ips/axis_cache.cpp
n-eiling: Hm I'm looked at how to make this work, but I'm really unhappy with your approach. The log loses so much value for me, if I don't log error where they occur. I mean in this example, it's definitely always an error if I set a register that does not exist.

How do you use logs if they don't tell you where errors occur? Maybe you need to explain to me how you use logs and how you determine where errors occur?
Using the debugger works well for this, even with exceptions, but e.g., in a real time application it's not always possible to use a debugger.

n-eiling:
i used the wrong mail address in the github ui. didn't know this affects the DCO check :)

fpga/lib/ips/axis_cache.cpp
stv0g: How about:

```cpp
define ExceptionWithLocation(...) Exception(__FILE__, __LINE__, __FUNCTION__, __VA_ARGS__)

function thrower() {
   # Bang
   throw ExceptionWithLocation("Bang, something happened here")
}

function catcher() {
   try {
      thrower();
   } catch (Exception &e) {
      logger->error(e)
   }
}
```

This would require two only overloads:

```cpp
Logger::error(Exception &e, fmt std::string, ...);
Exception::Exception(const char *file, int line, const char *function, fmt std::string, ...);
```

This approach has several advantages:
- We do not repeat error messages in the logger and in the exception
- We allow the `catcher()` to suppress the error log or reduce it severity (either because it may be ignored, or handled)
- We can make the logger configurable to output location information or not.

I understand that you as a developer want to see where the error occured.
But as a normal user this information is useless and just makes the output more noisy and harder to parse.

n-eiling: Yea I get that. And by the way, I already suggested this a while back. The static spdlog loggers do exactly this.

I still loose the order of logging outputs this way though. And I actually helped multiple people who where confused by exactly the resulting behavior: errors are at the very end despite everything looking good.

In my case it never happens that we want to ignore the error. And i think it would be bad design to use exception to implement logic.
The only way to make it okay to log in the exception handler would be if destructors never log anything to not confuse the log. This would probably be a good policy but currently node logs a lot it destructors.
Also we really don't have catch blocks. So the exception is often never handled.

In your suggestion: what is the point of log->error? Do you want it to allow it only at the top most level? 

By the way: if we go by popular guidelines, LLVM does not allow exception at all, the core guidelines say exceptions should only be used for error (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Re-errors), google also does not allow exception (https://google.github.io/styleguide/cppguide.html#Exceptions)

Anyway I would say we should move this discussion to the code guidelines issue. This is certainly not done consistently across the code base so if this uglyness here persists until we decided what to do, it shouldn't be too bad?

n-eiling: I removed the exception from AxisCache. Probably not what you wanted, but it resolves the strange pattern, without reducing the information value of logging.

n-eiling:
Let's merge this and move the discussion to https://github.com/VILLASframework/documentation/issues/89

n-eiling:
ping again.

PR #836, opened 2024-11-13
Add DCO.md

n-eiling:
We don't have a copy of the DCO in the repository. From my understanding we should have this to make it valid.

### DIFF
--- a/DCO.md
+++ b/DCO.md
@@ -1,0 +1,34 @@
+Developer Certificate of Origin
+Version 1.1
+
+Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
+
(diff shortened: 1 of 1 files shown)


n-eiling:
is there a way to disable the reuse check for this file?

stv0g:
Could you point me to the place where this requirement is mentioned?

I quickly tried to find the DCO in the Linux repo and could not find it.
Linux is one of the longest users of a DCO. So I would follow their style.

n-eiling:
Yes I was also looking for this. I'm not sure. I'll just link to the linux foundation copy in the contributor guidelines.

PR #839, opened 2024-11-21
Add a matcher for OS IP devices and IPs on a FPGA

IgnoreWarnings:
Match ip devices on os level with ips on the fpga based on memory address.

### DIFF
--- a/common/include/villas/kernel/devices/device_ip_matcher.hpp
+++ b/common/include/villas/kernel/devices/device_ip_matcher.hpp
@@ -1,5 +1,5 @@
 /* Matcher for IP devices in an OS devicetree and ips on a fpga board.
- * Matching based on adress prefix in device name and ip memoryaddress
+ * Matching based on address prefix in device name and ip memory address
  *
  * Author: Pascal Bauer <pascal.bauer@rwth-aachen.de>
  *


common/include/villas/kernel/devices/device_ip_matcher.hpp
```diff
@@ -0,0 +1,40 @@
+/* Matcher for IP devices in an OS devicetree and ips on a fpga board.
+ * Matching based on adress prefix in device name and ip memoryaddress
```
n-eiling, line 2:
```suggestion
 * Matching based on address prefix in device name and ip memory address
```

common/include/villas/kernel/devices/device_ip_matcher.hpp
```diff
@@ -0,0 +1,40 @@
...
+ * SPDX-FileCopyrightText: 2023-2024 Pascal Bauer <pascal.bauer@rwth-aachen.de>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <villas/fpga/core.hpp>
+#include <villas/kernel/devices/ip_device.hpp>
+
+class DeviceIpMatcher {
```
n-eiling, line 15: Why create a class? A function would be enough right?

n-eiling (changes requested):

PR #840, opened 2024-11-25
Added uldaq config to enable external clock source

matthiasnowak:
To allow the use of external clock sources with the node uldaq, a new config parameter called "sample_clock_source" has been added.
It can be either "internal" (behaving just as it used to before) or "external" (enabling the scan option "SO_EXTCLOCK").
Default behaviour (if the config is not set) is "internal".

### DIFF
--- a/lib/nodes/uldaq.cpp
+++ b/lib/nodes/uldaq.cpp
@@ -88,7 +88,7 @@ static const struct {
 static const struct {
   const char *name;
   ScanOption clock_source;
-} clock_sources[] = {{"internal", (ScanOption)( 0 << 4)},
+} clock_sources[] = {{"internal", (ScanOption)(0 << 4)},
                     {"external", SO_EXTCLOCK}};
 
 static AiInputMode uldaq_parse_input_mode(const char *str) {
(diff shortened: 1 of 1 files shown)


lib/nodes/uldaq.cpp
```diff
@@ -85,6 +85,12 @@ static const struct {
               {"unipolar-0.01", UNIPT01VOLTS, 0.0, +0.01},
               {"unipolar-0.005", UNIPT005VOLTS, 0.0, +0.005}};
 
+static const struct {
+  const char *name;
+  ScanOption clock_source;
+} clock_sources[] = {{"internal", (ScanOption)( 0 << 4)},
```
matthiasnowak, line 91: the uldaq library doesn't have a "select internal clock" option, this is supposed to "emulate" it. (could also be shortened to just 0)

matthiasnowak:
Could it be that the CI does not run on forks?

windrad6:
Looks good. Thank you

stv0g:
Thanks for the fixes. I've approved the PR. Please feel free to merge once the Pipleline succeeds

matthiasnowak:
> Thanks for the fixes. I've approved the PR. Please feel free to merge once the Pipleline succeeds

TYVM! I'm not sure if the pipeline will work since i'm working in a fork instead of a branch of the main repo. But the CI (as far as i understand it) works on a mirror of the main repo, right?

stv0g:
> But the CI (as far as i understand it) works on a mirror of the main repo, right?

Yes thats true :/

Do you know a good workaround?

PR #842, opened 2024-12-20
feat: Add basic tcp connection to socket node type

JitpanuM:
Add TCP option to socket node type

### DIFF
--- a/lib/nodes/socket.cpp
+++ b/lib/nodes/socket.cpp
@@ -102,8 +102,11 @@ char *villas::node::socket_print(NodeCompat *n) {
     break;
 
   case SocketLayer::TCP_SERVER:
+    layer = "tcp-server";
+    break;
+
   case SocketLayer::TCP_CLIENT:
-    layer = "tcp";
+    layer = "tcp-client";
     break;
   }
 
@@ -247,10 +250,10 @@ int villas::node::socket_start(NodeCompat *n) {
   }
 
   if (s->layer == SocketLayer::TCP_CLIENT) {
-    //Attempt to connect to TCP server
+    // Attempt to connect to TCP server.
     int retries = 0;
     while (retries < MAX_CONNECTION_RETRIES) {
-      n->logger->info("Attempting({}) to connect to server..", retries + 1);
+      n->logger->info("Attempting to connect to server: attempt={}...", retries + 1);
       ret = connect(s->sd, (struct sockaddr *)&s->out.saddr, addrlen);
       if (ret == 0) {
         break;
@@ -274,10 +277,10 @@ int villas::node::socket_start(NodeCompat *n) {
     }
   }
 
-  //TCP Server listen for client connection
+  // TCP Server listen for client connection.
   if (s->layer == SocketLayer::TCP_SERVER) {
     listen(s->sd, 5);
-    //Accept client connection and get client socket descriptor
+    // Accept client connection and get client socket descriptor.
     s->clt_sd = accept(s->sd, nullptr, nullptr);
     if (s->clt_sd < 0) {
       throw SystemError("Failed to accept connection");
@@ -365,7 +368,7 @@ int villas::node::socket_stop(NodeCompat *n) {
   }
 
   if (s->sd >= 0) {
-    //Close client socket descriptor
+    // Close client socket descriptor.
     if (s->layer == SocketLayer::TCP_SERVER)
       close(s->clt_sd);
 
@@ -396,10 +399,10 @@ int villas::node::socket_read(NodeCompat *n, struct Sample *const smps[],
   // Receive next sample
 
   if (s->layer == SocketLayer::TCP_CLIENT) {
-    //Receive data from server
+    // Receive data from server.
     bytes = recv(s->sd, s->in.buf, s->in.buflen, 0);
   } else if (s->layer == SocketLayer::TCP_SERVER) {
-    //Receive data from client
+    // Receive data from client.
     bytes = recv(s->clt_sd, s->in.buf, s->in.buflen, 0);
   } else {
     bytes = recvfrom(s->sd, s->in.buf, s->in.buflen, 0, &src.sa, &srclen);
@@ -510,10 +513,10 @@ retry:
 
 retry2:
   if (s->layer == SocketLayer::TCP_CLIENT) {
-    //Send data to TCP server
+    // Send data to TCP server.
     bytes = send(s->sd, s->out.buf, wbytes, 0);
   } else if (s->layer == SocketLayer::TCP_SERVER) {
-    //Send data to TCP client
+    // Send data to TCP client.
     bytes = send(s->clt_sd, s->out.buf, wbytes, 0);
   } else {
     bytes = sendto(s->sd, s->out.buf, wbytes, 0, (struct sockaddr *)&s->out.saddr,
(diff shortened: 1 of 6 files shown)


lib/nodes/socket.cpp
```diff
@@ -316,7 +365,12 @@ int villas::node::socket_stop(NodeCompat *n) {
   }
 
   if (s->sd >= 0) {
+    //Close client socket descriptor
```
stv0g, line 368:
```suggestion
    // Close client socket descriptor.
```

lib/nodes/socket.cpp
```diff
@@ -340,7 +394,17 @@ int villas::node::socket_read(NodeCompat *n, struct Sample *const smps[],
   socklen_t srclen = sizeof(src);
 
   // Receive next sample
-  bytes = recvfrom(s->sd, s->in.buf, s->in.buflen, 0, &src.sa, &srclen);
+
+  if (s->layer == SocketLayer::TCP_CLIENT) {
+    //Receive data from server
```
stv0g, line 399:
```suggestion
    // Receive data from server.
```

lib/nodes/socket.cpp
```diff
@@ -340,7 +394,17 @@ int villas::node::socket_read(NodeCompat *n, struct Sample *const smps[],
...
+    bytes = recv(s->sd, s->in.buf, s->in.buflen, 0);
+  } else if (s->layer == SocketLayer::TCP_SERVER) {
+    //Receive data from client
```
stv0g, line 402:
```suggestion
    // Receive data from client.
```

lib/nodes/socket.cpp
```diff
@@ -445,8 +509,17 @@ int villas::node::socket_write(NodeCompat *n, struct Sample *const smps[],
   }
 
 retry2:
-  bytes = sendto(s->sd, s->out.buf, wbytes, 0, (struct sockaddr *)&s->out.saddr,
+  if (s->layer == SocketLayer::TCP_CLIENT) {
+    //Send data to TCP server
```
stv0g, line 513:
```suggestion
    // Send data to TCP server.
```

lib/nodes/socket.cpp
```diff
@@ -445,8 +509,17 @@ int villas::node::socket_write(NodeCompat *n, struct Sample *const smps[],
...
+    bytes = send(s->sd, s->out.buf, wbytes, 0);
+  } else if (s->layer == SocketLayer::TCP_SERVER) {
+    //Send data to TCP client
```
stv0g, line 516:
```suggestion
    // Send data to TCP client.
```

lib/nodes/socket.cpp
```diff
@@ -505,6 +578,10 @@ int villas::node::socket_parse(NodeCompat *n, json_t *json) {
       s->layer = SocketLayer::UDP;
     else if (!strcmp(layer, "unix") || !strcmp(layer, "local"))
       s->layer = SocketLayer::UNIX;
+    else if (!strcmp(layer, "tcp_client"))
```
stv0g, line 581:
```suggestion
    else if (!strcmp(layer, "tcp-client"))
```

lib/nodes/socket.cpp
```diff
@@ -505,6 +578,10 @@ int villas::node::socket_parse(NodeCompat *n, json_t *json) {
...
+      s->layer = SocketLayer::TCP_CLIENT;
+    else if (!strcmp(layer, "tcp_server"))
```
stv0g, line 583:
```suggestion
    else if (!strcmp(layer, "tcp-server"))
```

lib/nodes/socket.cpp
```diff
@@ -97,6 +100,11 @@ char *villas::node::socket_print(NodeCompat *n) {
   case SocketLayer::UNIX:
     layer = "unix";
     break;
+
+  case SocketLayer::TCP_SERVER:
+  case SocketLayer::TCP_CLIENT:
+    layer = "tcp";
```
stv0g, line 106: Can we use `tcp-client` and `tcp-server` here as we do during parsing the options?

lib/nodes/socket.cpp
```diff
@@ -233,9 +246,43 @@ int villas::node::socket_start(NodeCompat *n) {
     addrlen = sizeof(s->in.saddr);
   }
 
-  ret = bind(s->sd, (struct sockaddr *)&s->in.saddr, addrlen);
-  if (ret < 0)
-    throw SystemError("Failed to bind socket");
+  if (s->layer == SocketLayer::TCP_CLIENT) {
+    //Attempt to connect to TCP server
```
stv0g, line 250:
```suggestion
    // Attempt to connect to TCP server.
```

lib/nodes/socket.cpp
```diff
@@ -233,9 +246,43 @@ int villas::node::socket_start(NodeCompat *n) {
...
+
+  if (ret < 0) {
+    if (s->layer == SocketLayer::TCP_CLIENT) {
+      throw SystemError("Failed to connect to TCP server");
+    } else {
+      throw SystemError("Failed to bind socket");
+    }
+  }
+
+  //TCP Server listen for client connection
```
stv0g, line 277:
```suggestion
  // TCP Server listen for client connection.
```

lib/nodes/socket.cpp
```diff
@@ -233,9 +246,43 @@ int villas::node::socket_start(NodeCompat *n) {
...
+  if (s->layer == SocketLayer::TCP_SERVER) {
+    listen(s->sd, 5);
+    //Accept client connection and get client socket descriptor
```
stv0g, line 280:
```suggestion
    // Accept client connection and get client socket descriptor.
```

lib/nodes/socket.cpp
```diff
@@ -233,9 +246,43 @@ int villas::node::socket_start(NodeCompat *n) {
...
   }
 
-  ret = bind(s->sd, (struct sockaddr *)&s->in.saddr, addrlen);
-  if (ret < 0)
-    throw SystemError("Failed to bind socket");
+  if (s->layer == SocketLayer::TCP_CLIENT) {
+    //Attempt to connect to TCP server
+    int retries = 0;
+    while (retries < MAX_CONNECTION_RETRIES) {
+      n->logger->info("Attempting({}) to connect to server..", retries + 1);
```
stv0g, line 253:
```suggestion
      n->logger->info("Attempting to connect to server: attempt={}...", retries + 1);
```

stv0g (commented):
Nice :) I like the PR 👍🏻 

Some open points:
- We need to update the OpenAPI spec of the socket node-type config here: https://github.com/VILLASframework/node/blob/master/doc/openapi/components/schemas/config/nodes/socket.yaml
- We need to extend the integration test here: https://github.com/VILLASframework/node/blob/master/tests/integration/node-loopback-socket.sh
- We need to update the documentation here: https://github.com/VILLASframework/documentation/blob/master/docs/node/nodes/socket.md
- We should be able to handle broken connections gracefully. Do we attempt to reconnect after the first successfully established connection?
- We might want to consider establishing the connection asynchronously, as doing so synchronously in `villas::node::socket_start(NodeCompat *n)` will block other nodes of the same VILLASnode instance to stall. In other words, VILLASnode will never finish its initialiazation until the TCP handshake has been completed..

PR #846, opened 2025-01-20
feat: TCP in node socket

al3xa23:
inlcude tcp in node-type socket

### DIFF
--- a/include/villas/nodes/socket.hpp
+++ b/include/villas/nodes/socket.hpp
@@ -24,6 +24,7 @@ struct Socket {
   int sd; // The socket descriptor
   int clt_sd; // TCP client socket descriptor
   int verify_source; // Verify the source address of incoming packets against socket::remote.
+  bool tcp_connect = false; // TCP connection status bit
 
   enum SocketLayer
       layer; // The OSI / IP layer which should be used for this socket
@@ -68,6 +69,8 @@ int socket_write(NodeCompat *n, struct Sample *const smps[], unsigned cnt);
 int socket_read(NodeCompat *n, struct Sample *const smps[], unsigned cnt);
 
 int socket_parse(NodeCompat *n, json_t *json);
+
+void socket_tcp_connection(NodeCompat *n, Socket *s);
 
 char *socket_print(NodeCompat *n);
 
--- a/include/villas/nodes/socket.hpp
+++ b/include/villas/nodes/socket.hpp
@@ -22,6 +22,7 @@ class NodeCompat;
 
 struct Socket {
   int sd; // The socket descriptor
+  int clt_sd; // TCP client socket descriptor
   int verify_source; // Verify the source address of incoming packets against socket::remote.
 
   enum SocketLayer
--- a/include/villas/nodes/socket.hpp
+++ b/include/villas/nodes/socket.hpp
@@ -24,7 +24,7 @@ struct Socket {
   int sd; // The socket descriptor
   int clt_sd; // TCP client socket descriptor
   int verify_source; // Verify the source address of incoming packets against socket::remote.
-  bool tcp_connect = false; // TCP connection status bit
+  bool tcp_connected = false; // TCP connection status bit
 
   enum SocketLayer
       layer; // The OSI / IP layer which should be used for this socket
@@ -69,8 +69,6 @@ int socket_write(NodeCompat *n, struct Sample *const smps[], unsigned cnt);
 int socket_read(NodeCompat *n, struct Sample *const smps[], unsigned cnt);
 
 int socket_parse(NodeCompat *n, json_t *json);
-
-void socket_tcp_connection(NodeCompat *n, Socket *s);
 
 char *socket_print(NodeCompat *n);
 
--- a/lib/nodes/socket.cpp
+++ b/lib/nodes/socket.cpp
@@ -102,8 +102,11 @@ char *villas::node::socket_print(NodeCompat *n) {
     break;
 
   case SocketLayer::TCP_SERVER:
+    layer = "tcp-server";
+    break;
+
   case SocketLayer::TCP_CLIENT:
-    layer = "tcp";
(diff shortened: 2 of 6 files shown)


include/villas/nodes/socket.hpp
```diff
@@ -22,7 +22,9 @@ class NodeCompat;
 
 struct Socket {
   int sd; // The socket descriptor
+  int clt_sd; // TCP client socket descriptor
   int verify_source; // Verify the source address of incoming packets against socket::remote.
+  bool tcp_connect = false; // TCP connection status bit
```
stv0g, line 27: The comment is not descriptive. What does this flag indicate? From the code, I can only guess that its used to keep track of the current connection state.

Can we please name this `tcp_connected` instead?

include/villas/nodes/socket.hpp
```diff
@@ -68,6 +70,8 @@ int socket_read(NodeCompat *n, struct Sample *const smps[], unsigned cnt);
 
 int socket_parse(NodeCompat *n, json_t *json);
 
+void socket_tcp_connection(NodeCompat *n, Socket *s);
```
stv0g, line 73: I think this is an internal function? Please remove it from the header, and declare it as `static` in the .cpp file.

tests/integration/node-loopback-socket.sh
```diff
@@ -87,3 +87,85 @@ wait %%
 # Send / Receive data to node
 VILLAS_LOG_PREFIX="[compare] " \
 villas compare input.dat output.dat
+
+sleep 1
```
stv0g, line 91: Why is this sleep required? Usually thats an indication of a bug in the code (race condition).

lib/nodes/socket.cpp
```diff
@@ -389,6 +436,39 @@ int villas::node::socket_read(NodeCompat *n, struct Sample *const smps[],
...
+        retries++;
+        if (retries < MAX_CONNECTION_RETRIES) {
+          sleep(RETRIES_DELAY);
+        }
+      }
+    }
+    if (ret < 0)
+      throw SystemError("Failed to conenct to TCP server");
+  } else if (s->layer == SocketLayer::TCP_SERVER) {
+    listen(s->sd, 5);
```
stv0g, line 462: Error checking?

lib/nodes/socket.cpp
```diff
@@ -389,6 +436,39 @@ int villas::node::socket_read(NodeCompat *n, struct Sample *const smps[],
   return ret;
 }
 
+void villas::node::socket_tcp_connection(NodeCompat *n, Socket *s) {
+  int ret;
+  if (s->layer == SocketLayer::TCP_CLIENT) {
+    close(s->sd);
```
stv0g, line 442: Error checking?

lib/nodes/socket.cpp
```diff
@@ -389,6 +436,39 @@ int villas::node::socket_read(NodeCompat *n, struct Sample *const smps[],
...
+    s->sd = socket(s->in.saddr.sa.sa_family, SOCK_STREAM, 0);
```
stv0g, line 443: Error checking?

lib/nodes/socket.cpp
```diff
@@ -389,6 +436,39 @@ int villas::node::socket_read(NodeCompat *n, struct Sample *const smps[],
...
+    // Attemp to connect to TCP server.
```
stv0g, line 444:
```suggestion
    // Attempt to connect to TCP server.
```

lib/nodes/socket.cpp
```diff
@@ -316,7 +338,12 @@ int villas::node::socket_stop(NodeCompat *n) {
   }
 
   if (s->sd >= 0) {
+    // Close client socket descriptor.
+    if (s->layer == SocketLayer::TCP_SERVER)
+      close(s->clt_sd);
```
stv0g, line 343: Error checking?

stv0g (changes requested):

lib/nodes/socket.cpp
```diff
@@ -316,7 +338,18 @@ int villas::node::socket_stop(NodeCompat *n) {
...
+    // Close client socket descriptor.
+    if (s->layer == SocketLayer::TCP_SERVER) {
+      close(s->clt_sd);
+      if (ret)
+        throw SystemError("Failed to close TCP client socket descriptor");
+    }
+
     ret = close(s->sd);
+
+    //Reset socket descriptor
```
stv0g, line 350:
```suggestion
    // Reset socket descriptor.
```

stv0g (changes requested):

tests/integration/node-loopback-socket.sh
JitpanuM: This sleep is not required as you said. So, I removed it from the test.

include/villas/nodes/socket.hpp
JitpanuM: It is an internal function. 
I have declared it as static as you suggested.

lib/nodes/socket.cpp
JitpanuM: My mistake, I have added the error case for every function you suggested.

include/villas/nodes/socket.hpp
stv0g: Thank you :) 👍🏻

lib/nodes/socket.cpp
stv0g: Thanks :)

lib/nodes/socket.cpp
stv0g:
```suggestion
      ret = close(s->clt_sd);
```

stv0g (changes requested):
Nice work @JitpanuM,

I've rebased your changes to the latest master to trigger the CI again.

There is only a last missing/broken error check.

PR #845, opened 2025-01-16
Add SMU node-type

al3xa23:
Created a new smu branch because commit history was pretty messed up. Everything now is based on the status of the the old branch, but all smu changes included in one commit. Hope it is fine like that

PR #849, opened 2025-01-24
chore: remove catchall codeowner

n-eiling:
@stv0g You as default code owner for everything just doesn't work anymore.
We cannot wait for 3 months for a review. I'd suggest you add yourself to parts that are very important to you.
Or do you have another idea for a solution?

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,12 +1,6 @@
 # SPDX-FileCopyrightText: 2014-2023 Institute for Automation of Complex Power Systems, RWTH Aachen University
 # SPDX-License-Identifier: Apache-2.0
 
-# These owners will be the default owners for everything in
-# the repo. Unless a later match takes precedence,
-# @stv0g will be requested for
-# review when someone opens a pull request.
-*       @stv0g
-
 # Order is important; the last matching pattern takes the most
 # precedence. When someone opens a pull request that only
 # modifies the following files, only the following members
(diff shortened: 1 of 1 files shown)


stv0g:
> We cannot wait for 3 months for a review.

I agree. It does not work with me as the only reviewer and maintainer.

> Or do you have another idea for a solution?

So, in my opinion the solution should not be to remove me as a codeowner, but instead find other people who are interested in this task.

Do we have any volunteers?

Unfortunately, there seems nobody else interested in providing some reviews. And by "reviewing" I dont mean to simply "approve" a PR from a colleague to do them a favour.
A review takes time and some effort to understand the changes. So a review without any comments, questions or remarks is not proper review.

It like with publications..

It is a bit sad :( I am against removing the requirement for a review as this will deteriorate the code quality over time. If you we dont find somebody else to work on this, I suggest we better fork personal work into personal repos or split the project again.

stv0g:
@n-eiling Are there more PR's to review? I provided quite a few reviews over the last weeks (see https://github.com/VILLASframework/node/pull/853, https://github.com/VILLASframework/node/pull/846, https://github.com/VILLASframework/node/pull/845). But those reviews have remained all unanswered which is not improving my motivation to dedicate time to this :/

n-eiling:
I don't have a good solution to all of this.
We will not find anyone willing to t\do code reviews on everything. 
Code quality will deteriorate.
But if we don't change anything, the project will not get any more contributions.

stv0g:
@n-eiling Would you volunteer as a second catch-all codeowner?

stv0g:
@n-eiling ping

n-eiling:
Yes, but I don't think this will help much. I will not have a lot of time to do reviews in the future. We should find even more people.

stv0g (approved):

PR #854, opened 2025-01-29
feat: Vfio platform interrupts

IgnoreWarnings:
Add a method to enable interrupts on platform.


stv0g (approved):
Looks good to me.

It would be nice to be a bit more consistent with new-lines. But thats a nit.

n-eiling (changes requested):
Please add checking for the AUTOMASK flag. Maybe we can abstract the VFIO interrupt code into a vfioInterruptController super class?
Both platform and pcie interrupt controllers should use the same code for waitForInterrupt, with the difference that the pcie interrupt controller also supports poling the interrupt controller.

PR #859, opened 2025-02-01
Fixes to compile in Ampere processor & Ubuntu

leonardocarreras:
This PR has the following goals:
- Compile using Ubuntu 24.04
- Compile in ARM64 architecture (aarch64).

**Changelog**
- On the script for dependencies: 
    - Repository updated for redis++, version bump 1.3.7 -> 1.3.8 to fix compilation issue.
    - Failure for distutils, installing now setuptools in the virtual environment for compilation of libnice.
- In the documentation and docker container for Ubuntu:
    -  Add dependencies needed for libnice and libdatachannel compilation.
 - Timestamp counter (TSC) and tests for Kernel and Queue updated for compatibility:
     - Ensured nop() function works correctly on both x86 and ARM. Replaced rep nop with yield on ARM (__aarch64__) to ensure cross-platform compatibility. 
     - Adapt tsc_init to support compatibility and obtain data from the kernel call as a fallback case.
     - Move the tsc.cpp file to the linux-required section of CMakeLists.

### DIFF
--- a/common/include/villas/tsc.hpp
+++ b/common/include/villas/tsc.hpp
@@ -32,7 +32,7 @@
 #endif
 
 #ifndef CLOCK_MONOTONIC_RAW
-#define CLOCK_MONOTONIC_RAW CLOCK_MONOTONIC // Fallback for macOS/BSD
+#define CLOCK_MONOTONIC_RAW CLOCK_MONOTONIC // Fallback for Darwin/BSD
 #endif
 
 #endif
@@ -51,7 +51,7 @@ __attribute__((unused)) static uint64_t tsc_now(struct Tsc *t) {
   return t->rdtscp_supported ? __rdtscp(&tsc_aux) : __rdtsc();
 }
 #else
-// Fallback for CLOCK_MONOTONIC_RAW (linux/BSD/mac specific)
+// Fallback for CLOCK_MONOTONIC_RAW (Linux/Darwin/BSD specific)
 __attribute__((unused)) static uint64_t tsc_now(struct Tsc *t) {
   struct timespec ts;
   clock_gettime(CLOCK_MONOTONIC_RAW, &ts);
--- a/common/include/villas/tsc.hpp
+++ b/common/include/villas/tsc.hpp
@@ -1,4 +1,4 @@
-/* Measure time and sleep with IA-32 time-stamp counter.
+/* Measure time and sleep with IA-32 time-stamp counter (x86) or a generic cycle-based timing.
  *
  * Author: Steffen Vogel <post@steffenvogel.de>
  * SPDX-FileCopyrightText: 2014-2023 Institute for Automation of Complex Power Systems, RWTH Aachen University
@@ -7,11 +7,10 @@
 
 #pragma once
 
-#if !(__x86_64__ || __i386__)
-#error this header is for x86 only
-#endif
-
 #include <cinttypes>
+
+#if defined(__x86_64__) || defined(__i386__)
+
 #include <cpuid.h>
 #include <x86intrin.h>
 
@@ -24,17 +23,41 @@
 #define bit_TSC_INVARIANT (1 << 8)
 #define bit_RDTSCP (1 << 27)
 
+#else
+
+#include <time.h>
+
+#ifndef NSEC_PER_SEC
+#define NSEC_PER_SEC 1000000000L
+#endif
+
+#ifndef CLOCK_MONOTONIC_RAW
+#define CLOCK_MONOTONIC_RAW CLOCK_MONOTONIC // Fallback for macOS/BSD
+#endif
+
+#endif
+
 struct Tsc {
   uint64_t frequency;
-
+#if defined(__x86_64__) || defined(__i386__)
   bool rdtscp_supported;
   bool is_invariant;
+#endif
 };
 
+#if defined(__x86_64__) || defined(__i386__)
 __attribute__((unused)) static uint64_t tsc_now(struct Tsc *t) {
   uint32_t tsc_aux;
   return t->rdtscp_supported ? __rdtscp(&tsc_aux) : __rdtsc();
 }
+#else
+// Fallback for CLOCK_MONOTONIC_RAW (linux/BSD/mac specific)
+__attribute__((unused)) static uint64_t tsc_now(struct Tsc *t) {
+  struct timespec ts;
+  clock_gettime(CLOCK_MONOTONIC_RAW, &ts);
+  return (ts.tv_sec * NSEC_PER_SEC + ts.tv_nsec) * t->frequency / NSEC_PER_SEC;
+}
+#endif
 
 int tsc_init(struct Tsc *t) __attribute__((warn_unused_result));
 
--- a/common/lib/tsc.cpp
+++ b/common/lib/tsc.cpp
@@ -7,9 +7,8 @@
 
 #include <villas/tsc.hpp>
 
-using namespace villas;
-
 int tsc_init(struct Tsc *t) {
+#if defined(__x86_64__) || defined(__i386__)
   uint32_t eax = 0, ebx = 0, ecx = 0, edx = 0;
 
   // Check if TSC is supported
@@ -32,15 +31,25 @@ int tsc_init(struct Tsc *t) {
   if (ecx != 0)
     t->frequency = ecx * ebx / eax;
   else {
-    int ret;
 #ifdef __linux__
-    ret = kernel::get_cpu_frequency(&t->frequency);
+    int ret = kernel::get_cpu_frequency(&t->frequency);
     if (ret)
       return ret;
 #endif
   }
-
-  return 0;
+#elif defined(__aarch64__)
+  // Read counter frequency from system register
+  uint64_t cntfrq;
+  asm volatile("mrs %0, cntfrq_el0" : "=r"(cntfrq));
+  t->frequency = cntfrq;
+#else
+#ifdef __linux__
+  int ret = kernel::get_cpu_frequency(&t->frequency);
+  if (ret)
+    return ret;
+#endif
+#endif
+  return (int)(t->frequency);
 }
 
 uint64_t tsc_rate_to_cycles(struct Tsc *t, double rate) {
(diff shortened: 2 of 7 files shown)


common/include/villas/tsc.hpp
```diff
@@ -7,14 +7,16 @@
...
-#error this header is for x86 only
-#endif
-
 #include <cinttypes>
+
+#if defined(__x86_64__) || defined(__i386__)
+
+// ==========================
+// x86 IMPLEMENTATION (TSC)
+// ==========================
```
stv0g, line 16:
```suggestion
// x86 Implementation (TSC)
```

common/include/villas/tsc.hpp
```diff
@@ -7,14 +7,16 @@
...
+
 #include <cpuid.h>
 #include <x86intrin.h>
-
```
stv0g, line 17: Please separate the system headers from our headers with an empty line.

common/include/villas/tsc.hpp
```diff
@@ -25,17 +27,85 @@
 #define bit_RDTSCP (1 << 27)
 
 struct Tsc {
-  uint64_t frequency;
-
-  bool rdtscp_supported;
-  bool is_invariant;
+    uint64_t frequency;
```
stv0g, line 30: Did you format your code with clang-format? The indentation seems to have changed here.

common/include/villas/tsc.hpp
```diff
@@ -25,17 +27,85 @@
...
+int tsc_init(struct Tsc *t);
+
 uint64_t tsc_rate_to_cycles(struct Tsc *t, double rate);
+
+#elif defined(__aarch64__)
+
+// ==========================
+// ARM (AArch64) IMPLEMENTATION
+// Uses CLOCK_MONOTONIC_RAW but returns cycles instead of nanoseconds
+// ==========================
```
stv0g, line 53:
```suggestion
// ARM (AArch64) Implementation
// Uses CLOCK_MONOTONIC_RAW but returns cycles instead of nanoseconds
```

common/include/villas/tsc.hpp
```diff
@@ -25,17 +27,85 @@
...
+
+#include <time.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+// Define a constant for nanoseconds per second
```
stv0g, line 59:
```suggestion
// Define a constant for nanoseconds per second.
```

common/include/villas/tsc.hpp
```diff
@@ -25,17 +27,85 @@
...
+#define NANOSECONDS_PER_SECOND 1000000000ULL
+
+struct Tsc {
+    uint64_t frequency; // Must be initialized based on hardware
```
stv0g, line 63:
```suggestion
    uint64_t frequency; // Must be initialized based on hardware.
```

common/include/villas/tsc.hpp
```diff
@@ -25,17 +27,85 @@
...
+    bool rdtscp_supported;
+    bool is_invariant;
+};
+
+// Gets the current time in cycles, similar to x86 TSC
```
stv0g, line 68:
```suggestion
// Gets the current time in cycles, similar to x86 TSC.
```

common/include/villas/tsc.hpp
stv0g: As far as I can see there is nothing specific to ARM by using `clock_get_time()`.

Can we use this implementation for all non-x86 architectures?

common/include/villas/tsc.hpp
```diff
@@ -25,17 +27,85 @@
...
+// Convert rate to cycles based on CPU frequency
+uint64_t tsc_rate_to_cycles(struct Tsc *t, double rate) {
+    return (uint64_t)(rate * t->frequency);
+}
+
+#else
+
+// ==========================
+// UNSUPPORTED ARCHITECTURE
+// ==========================
```
stv0g, line 107:
```suggestion
```

Please dont duplicate the `#error` from below. This is just stuff we need to additionally maintain.

packaging/deps.sh
stv0g, line 1: Could you elaborate, why we need to bump the versions here?

Did some minimal version requirements change?

packaging/deps.sh
```diff
@@ -426,12 +426,13 @@ if ! pkg-config "nice >= 0.1.16" && \
             . venv/bin/activate
 
             # Note: meson 0.61.5 is the latest version which supports the CMake version on the target
-            pip3 install meson==0.61.5
+            pip3 install meson
```
stv0g, line 429: Why are we no-longer pinning the meson version?

stv0g (changes requested):
- CI tests are somehow not triggered?
- Please rebase on master

packaging/deps.sh
leonardocarreras: some of this bumps were due compiling in the newest versions of the compilers and packages, I tried now with a VM as-it-is with Ubuntu 24.04 (aarch64-2024.10.09-0) without making big changes, and for now:
- redis++ does not compile (errors in utils.h)  => minimum version that compiles is 1.3.8
- datachannel does not compile (linker error in tests) => error with glib (different requirements between libnice and libdatachannel)

I will look into it a bit more before giving a conclusion, but libnice uses a [subproject](https://gitlab.freedesktop.org/libnice/libnice/-/blob/master/subprojects/glib.wrap?ref_type=heads) with v2.64 and for ARM compilation libdatachannel needs functions that are available in glib v2.76+

common/include/villas/tsc.hpp
leonardocarreras: I think we can, I will change this

common/include/villas/tsc.hpp
leonardocarreras: Will fix in the next rebase + adjustments

packaging/deps.sh
leonardocarreras: Most of this will change back (investigating now the problem of libnice vs libdatachannel)

packaging/deps.sh
leonardocarreras: we need to install first `python3-venv libglib2.0-0 libglib2.0-dev`, I will add this to the help

leonardocarreras: I understand that would be in the table:
https://github.com/VILLASframework/documentation/blob/42728147973f9a57888b7e1a2971eceac848720b/docs/node/installation.md?plain=1#L66

In this installation for Ubuntu/Debian
https://github.com/VILLASframework/documentation/blob/42728147973f9a57888b7e1a2971eceac848720b/docs/node/installation.md#L104

And the RHEL/Fedora & co
https://github.com/VILLASframework/documentation/blob/42728147973f9a57888b7e1a2971eceac848720b/docs/node/installation.md#L141

Does this follow any specific order?

common/include/villas/tsc.hpp
leonardocarreras: Used the time library constant

leonardocarreras:
The error with cppcheck in the CI [here](https://git.rwth-aachen.de/acs/public/villas/node/-/jobs/5859046#L66 )
image
Is in a line I did not touch
https://github.com/leonardocarreras/node/blob/ccda404460721d927a7810d8c81c198928ae1890/tests/unit/queue.cpp#L208

stv0g:
> Should I use `// cppcheck-suppress unknownMacro` to make the CI pass?

Yes, this is a macro defined by the unit test framework which we use.

common/lib/tsc.cpp
```diff
@@ -32,15 +34,28 @@ int tsc_init(struct Tsc *t) {
...
     if (ret)
       return ret;
 #endif
   }
-
-  return 0;
+#elif defined(__aarch64__)
+  // Read counter frequency from system register
+  uint64_t cntfrq;
+  asm volatile("mrs %0, cntfrq_el0" : "=r"(cntfrq));
```
stv0g, line 46: Is this really required? I think we can use the same SysFS based approach (`kernel::get_cpu_frequency()`). It should be already portable and allows us the get rid of this special case.

packaging/deps.sh
stv0g: >  should I add this to the help? 

Yes that would be great.

> Does this follow any specific order?

None that I am aware of. If you change it, please feel free to simply sort them alphabetically.

stv0g (changes requested):
Looks great :)

I have a last request: could you update the included Ubuntu Dockerfile to 24.4 as well?

See: https://github.com/VILLASframework/node/blob/master/packaging/docker/Dockerfile.ubuntu

common/lib/tsc.cpp
leonardocarreras: Well, probably not really but I followed the approach of the existing Intel function, I am taking this part out

leonardocarreras:
> >  should I add this to the help?
> 
> Yes that would be great.
> 
> >  Does this follow any specific order?
> 
> None that I am aware of. If you change it, please feel free to simply sort them alphabetically.

PR sent, see https://github.com/VILLASframework/documentation/pull/94

common/include/villas/tsc.hpp
```diff
@@ -24,17 +23,41 @@
...
+#endif
 };
 
+#if defined(__x86_64__) || defined(__i386__)
 __attribute__((unused)) static uint64_t tsc_now(struct Tsc *t) {
   uint32_t tsc_aux;
   return t->rdtscp_supported ? __rdtscp(&tsc_aux) : __rdtsc();
 }
+#else
+// Fallback for CLOCK_MONOTONIC_RAW (linux/BSD/mac specific)
```
stv0g, line 54:
```suggestion
// Fallback for CLOCK_MONOTONIC_RAW (Linux/Darwin/BSD specific)
```

packaging/docker/Dockerfile.ubuntu
```diff
@@ -53,7 +56,10 @@ RUN apt-get update && \
 		liblua5.3-dev \
 		libhiredis-dev \
 		libnice-dev \
-		libmodbus-dev
+		libmodbus-dev \
+		libre2-dev \
+		libglib2.0-0 libglib2.0-dev \
```
stv0g, line 61:
```suggestion
		libglib2.0-dev \
```

stv0g (changes requested):

stv0g (approved):

PR #860, opened 2025-02-07
feat: Vfio Device connection

IgnoreWarnings:
Add a connection to devicetree devices via vfio.

### DIFF
--- a/common/include/villas/kernel/devices/device_connection.hpp
+++ b/common/include/villas/kernel/devices/device_connection.hpp
@@ -1,0 +1,79 @@
+/* Vfio connection to a device.
+ *
+ * Author: Pascal Bauer <pascal.bauer@rwth-aachen.de>
+ *
+ * SPDX-FileCopyrightText: 2024-25 Pascal Bauer <pascal.bauer@rwth-aachen.de>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <memory>
+
+#include <villas/kernel/vfio_device.hpp>
+#include <villas/kernel/devices/linux_driver.hpp>
+
+namespace villas {
+namespace kernel {
+namespace devices {
+
+class DeviceConnection {
+public:
+  Logger logger;
+  const std::shared_ptr<kernel::vfio::Device> vfio_device;
+
+private:
+  DeviceConnection(std::shared_ptr<kernel::vfio::Device> vfio_device)
+      : logger(villas::Log::get("DeviceConnection")),
+        vfio_device(vfio_device){};
+
+public:
+  static DeviceConnection
+  from(const villas::kernel::devices::Device& device,
+       std::shared_ptr<kernel::vfio::Container> vfio_container) {
+    auto logger = villas::Log::get("Builder: DeviceConnection");
+
+    // Bind to driver
+    LinuxDriver driver(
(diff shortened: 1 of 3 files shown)


common/include/villas/kernel/devices/device_connection.hpp
```diff
@@ -0,0 +1,79 @@
...
+    auto vfio_device = std::make_shared<kernel::vfio::Device>(
+        device.name(), vfio_group->getFileDescriptor());
+
+    // Attach device to group
+    vfio_group->attachDevice(vfio_device);
+
+    return DeviceConnection(vfio_device);
+  }
+
+  void add_to_memorygraph() const {
```
stv0g, line 56: Please note that the LLVM code-style is mandating camelCase for function names:

https://releases.llvm.org/10.0.0/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly

common/include/villas/kernel/devices/device_connection.hpp
```diff
@@ -0,0 +1,79 @@
...
+private:
+  DeviceConnection(std::shared_ptr<kernel::vfio::Device> vfio_device)
+      : logger(villas::Log::get("DeviceConnection")),
+        vfio_device(vfio_device) {};
+
+public:
+  static DeviceConnection
+  from(const villas::kernel::devices::Device &device,
+       std::shared_ptr<kernel::vfio::Container> vfio_container) {
+    auto logger = villas::Log::get("Builder: DeviceConnection");
```
stv0g, line 34: Please move code to the `.cpp` file.

common/include/villas/kernel/devices/device_connection.hpp
```diff
@@ -0,0 +1,79 @@
...
+#include <memory>
+
+#include <villas/kernel/devices/linux_driver.hpp>
+#include <villas/kernel/vfio_device.hpp>
+
+namespace villas {
+namespace kernel {
+namespace devices {
+
+class DeviceConnection {
```
stv0g, line 20: Could you motivate why we need yet another class to manage devices and VFIO groups, containers?

I already lost track on how many classes we have 😁 

It would be good to understand why they are needed, as it looks like this class in not actually containing the code to bind a device to a VFIO container/group, but instead refers to the VFIO container/group classes?

stv0g (changes requested):

common/include/villas/kernel/devices/device_connection.hpp
n-eiling: A lot of this is motivated by not breaking the PCIe code. @IgnoreWarnings  can you check if this cannot be integrated into Device?

n-eiling:
can you please push to gitlab to make the CI happy?

n-eiling (approved):

common/include/villas/kernel/devices/device_connection.hpp
stv0g: The VFIO API was designed to abstract IOMMU mappings across different types of devices. E.g. platform and PCIe. Hence, the API for attaching devices to VFIO groups and containers is fairly similar.

In my opinion, we should not make it more complicated by implementing this separately for PCIe and Platform devices as this would just mean we lose the generalisation which the Kernel VFIO sub-system is providing us.

PR #867, opened 2025-02-26
refactor: move singular vfio device requirement to PCIeCard

IgnoreWarnings:
Pcie card works with exactly one vfio device.
The new improved design for platform cards will any number of vfio devices.

### DIFF
--- a/fpga/include/villas/fpga/card.hpp
+++ b/fpga/include/villas/fpga/card.hpp
@@ -25,7 +25,6 @@ public:
 
   std::string name; // The name of the FPGA card
   std::shared_ptr<kernel::vfio::Container> vfioContainer;
-  std::shared_ptr<kernel::vfio::Device> vfioDevice;
 
   // Slave address space ID to access the PCIe address space from the
   // FPGA
--- a/fpga/lib/ips/intc.cpp
+++ b/fpga/lib/ips/intc.cpp
@@ -22,17 +22,20 @@ using namespace villas::fpga::ip;
 InterruptController::~InterruptController() {}
 
 bool InterruptController::stop() {
-  return card->vfioDevice->pciMsiDeinit(this->efds) > 0;
+  return this->vfioDevice->pciMsiDeinit(this->efds) > 0;
 }
 
 bool InterruptController::init() {
   const uintptr_t base = getBaseAddr(registerMemory);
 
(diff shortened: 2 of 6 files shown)


fpga/include/villas/fpga/card.hpp
```diff
@@ -25,7 +25,6 @@ class Card {
 
   std::string name; // The name of the FPGA card
   std::shared_ptr<kernel::vfio::Container> vfioContainer;
-  std::shared_ptr<kernel::vfio::Device> vfioDevice;
```
stv0g, line 28: Could you please add a description to the PR? Why is the vfioDevice handle something speciic to PCIe cards? In my understanding VFIO is supposed to abstract beyond PCIe and also applies to Platform devices.

fpga/lib/ips/intc.cpp
```diff
@@ -22,17 +22,20 @@ using namespace villas::fpga::ip;
 InterruptController::~InterruptController() {}
 
 bool InterruptController::stop() {
-  return card->vfioDevice->pciMsiDeinit(this->efds) > 0;
+  return this->vfioDevice->pciMsiDeinit(this->efds) > 0;
```
n-eiling, line 25: `this` is not needed here

fpga/lib/ips/intc.cpp
```diff
@@ -22,17 +22,20 @@ using namespace villas::fpga::ip;
...
-  num_irqs = card->vfioDevice->pciMsiInit(efds);
+  PCIeCard *pciecard = dynamic_cast<PCIeCard *>(card);
+  this->vfioDevice = pciecard->vfioDevice;
+
+  num_irqs = this->vfioDevice->pciMsiInit(efds);
   if (num_irqs < 0)
     return false;
 
-  if (not card->vfioDevice->pciMsiFind(nos)) {
+  if (not this->vfioDevice->pciMsiFind(nos)) {
```
n-eiling, line 38: you can also remove all the `this` here

n-eiling (changes requested):

fpga/lib/ips/intc.cpp
n-eiling: no it's fine with me.

n-eiling (approved):

PR #869, opened 2025-03-10
Fix: fpga wrong dino ip name, move do reset of card to parent

IgnoreWarnings:
The ip name of dinoif_fast was corrected to dinoif_adc.
The member doReset of pciecard was moved to parent class card.


n-eiling:
can you push to gitlab? it should merge automatically once the pipeline suceeds.

PR #876, opened 2025-03-17
Python-Wrapper commits

KeVteL:
### Left to do:

Add docstrings to the python-wrapper bindings once documentation is finalized.
 
### Side notes:

As noted in a64e0dc, potential CI issue with running the python-wrapper tests (permissions).

sample_pack() and sample_unpack() may need additional work, not yet tested.

### Following could be investigated:

node_name_short() bug - appears to be returning a free'd string.

node_stop() and node_restart() do not work as expected with the signal_v2 node in the dev container.
Behavior between some functions differs from my live system to the dev container.

As an example:
node_check() and node_prepare() do not need to be called before starting the signal_v2 node (works just fine on live system).
node_stop() and node_restart() work just fine as well.

That however is not the case in the dev container.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -170,6 +170,7 @@ test:unit:
   script:
     - cmake -S . -B build ${CMAKE_OPTS}
     - cmake --build build ${CMAKE_BUILD_OPTS} --target run-unit-tests run-unit-tests-common
+    - cmake --build build ${CMAKE_BUILD_OPTS} --target python-wrapper-tests
   needs:
     - job: "build:source: [fedora]"
       artifacts: true
--- a/clients/python-wrapper/CMakeLists.txt
+++ b/clients/python-wrapper/CMakeLists.txt
@@ -1,3 +1,9 @@
+# Python-wrapper CMakeLists.
+#
+# Author: Kevin Vu te Laar <vu.te@rwth-aachen.de>
+# SPDX-FileCopyrightText: 2014-2025 Institute for Automation of Complex Power Systems, RWTH Aachen University
+# SPDX-License-Identifier: Apache-2.0
+
 cmake_minimum_required(VERSION 3.15...3.29)
 project(villas-python-wrapper LANGUAGES CXX)
 
--- a/clients/python-wrapper/CMakeLists.txt
+++ b/clients/python-wrapper/CMakeLists.txt
@@ -1,0 +1,32 @@
+cmake_minimum_required(VERSION 3.15...3.29)
+project(villas-python-wrapper LANGUAGES CXX)
+
+set(PYBIND11_FINDPYTHON ON)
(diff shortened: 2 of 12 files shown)


clients/python-wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+cmake_minimum_required(VERSION 3.15...3.29)
```
stv0g, line 1: Please add SPDX headers as in other files.

clients/python-wrapper/villas-python-wrapper.cpp
```diff
@@ -0,0 +1,235 @@
+#include <jansson.h>
```
stv0g, line 1: Please add SPDX headers as in other files.

clients/python-wrapper/villas-python-wrapper.cpp
```diff
@@ -0,0 +1,235 @@
...
+#include <pybind11/pybind11.h>
+#include <uuid/uuid.h>
+#include <villas/node.hpp>
+#include <villas/sample.hpp>
+extern "C" {
```
stv0g, line 6:
```suggestion

extern "C" {
```

clients/python-wrapper/villas-python-wrapper.cpp
```diff
@@ -0,0 +1,235 @@
...
+    unsigned int size() const {
+      return len;
+    }
+
+  private:
+    vsample **smps;
+    unsigned int len;
+};
+
+//pybind11 can not deal with (void **) as function input parameters,
```
stv0g, line 47: Please add spaces between `//` and the following text:

```suggestion
// pybind11 can not deal with (void **) as function input parameters,
```

packaging/docker/Dockerfile.rocky
```diff
@@ -25,7 +25,8 @@ RUN dnf --allowerasing -y install \
 	flex bison \
 	texinfo git curl tar \
 	protobuf-compiler protobuf-c-compiler \
-	clang-tools-extra
+	clang-tools-extra \
+  python python-devel
```
stv0g, line 29:
```suggestion
	python python-devel
```

tests/unit/python_wrapper.py
```diff
@@ -0,0 +1,295 @@
+import json
```
stv0g, line 1: Please add SPDX file header as in other Python files.

stv0g (changes requested):
Nice work :)

I've added some initial comments. I know the PR is still in draft state, but I thought it would help us to merge this more quickly.

.gitlab-ci.yml
KeVteL, line 1: An additional
`cmake --build build ${CMAKE_BUILD_OPTS} --target ...`
call, as python-wrapper-tests has differing dependencies from run-unit-tests and run-unit-tests-common.

PR #877, opened 2025-03-18
feat: Add a CoreConnection to connect fpga Ip cores to the memory graph

IgnoreWarnings:

PR #878, opened 2025-03-18
feat: Add a Platform card class to interface with platform based FPGAs

IgnoreWarnings:
rebased on #877.


n-eiling (approved):
I think the changes are

stv0g:
Please use more descriptive commit messages. A two word description like `add parser` is not sufficient.

@n-eiling

PR #879, opened 2025-03-20
Add miob example config

IgnoreWarnings:
Add a miob example config for loopback.

### DIFF
--- a/tests/integration/api-nodes.sh
+++ b/tests/integration/api-nodes.sh
@@ -25,29 +25,29 @@ cat > config.json <<EOF
     },
     "nodes": {
         "testnode1": {
-             "type": "websocket",
-             "dummy": "value1"
+            "type": "websocket",
+            "dummy": "value1"
         },
         "testnode2": {
-             "type": "socket",
-             "dummy": "value2",
+            "type": "socket",
+            "dummy": "value2",
 
-             "in": {
-             	"address": "*:12001",
-             	"signals": [
-             		{ "name": "sig1", "unit": "Volts",  "type": "float", "init": 123.0 },
-             		{ "name": "sig2", "unit": "Ampere", "type": "integer", "init": 123 }
-             	]
-             },
-             "out": {
-             	"address": "127.0.0.1:12000"
-             }
+            "in": {
+                "address": "*:12001",
+                "signals": [
+                    { "name": "sig1", "unit": "Volts",  "type": "float", "init": 123.0 },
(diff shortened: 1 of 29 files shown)


n-eiling:
please name the file fpga-miob. The filename should reference the node type the file is an example for.

stv0g:
The new config fails in the integration test:

```
=== Testing config: /builds/acs/public/villas/node/etc/examples/nodes/fpga-miob.conf
17:11:54 info memory           Initialize memory sub-system: #hugepages=100
17:11:54 warn memory:mmap      Running in an unprivileged environment. Hugepages are not used!
17:11:54 warn memory           Running in an unprivileged environment. Memory is not locked to RAM!
17:11:54 info test-config      This is VILLASnode v0.11.0-5a24f6b-release-ci (built on Mar 21 2025, 17:10:18)
17:11:54 info signals          Initialize subsystem
17:11:54 info config           Reading configuration from local file: /builds/acs/public/villas/node/etc/examples/nodes/fpga-miob.conf
modprobe: FATAL: Module vfio not found in directory /lib/modules/5.15.0-1073-kvm
17:11:54  err test-config      Kernel module 'vfio' required but could not be loaded. Please load manually!
[FAIL]  test-config (exited with code 255, ran for 2.366331618s)
Summary: 1 of 61 tests failed.
   Timedout: 0
   Skipped: 18
gmake[3]: *** [tests/integration/CMakeFiles/run-integration-tests.dir/build.make:73: tests/integration/CMakeFiles/run-integration-tests] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:1506: tests/integration/CMakeFiles/run-integration-tests.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:1513: tests/integration/CMakeFiles/run-integration-tests.dir/rule] Error 2
gmake: *** [Makefile:631: run-integration-tests] Error 2
```

n-eiling:
fpga test will always fail because the CI does not have a FPGA.

stv0g:
The test will only attempt to parse the configuration. No actual exchange via VFIO needs to take place here. The point is only to verify that the configuration file is semantically correct and can be parsed.

n-eiling:
the parsing shouldn't try to load the vfio kernel module. @IgnoreWarnings can you check what is happening? Maybe you have to separate the device resolution from the parsing.

stv0g:
@IgnoreWarnings I see you removed the module loading for `vfio_platform`. The test is now failing on another module: `vfio`

PR #882, opened 2025-04-02
chore(node-iec61850): Upgrade libiec61850 to 1.6.0

stv0g:


n-eiling:
@windrad6 are you using this?

windrad6:
> @windrad6 are you using this?

Yes we want to use R-GOOSE soon

PR #884, opened 2025-04-04
Add Python bindings for libvillas

al3xa23:
Python Wrapper for VILLASnode 
Documentation available https://github.com/VILLASframework/documentation/pull/104#issue-2971440658 
Please provide comments what needs to be changed from you point of view :)


tests/unit/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ add_custom_target(run-unit-tests
     USES_TERMINAL
 )
 
+find_package(Python3 REQUIRED COMPONENTS Interpreter)
```
stv0g, line 38: Are you sure this is needed? I dont see the variables/targets defined by `find_package` being used in this file.

Normally, such transitive dependencies should be defined using the `PUBLIC` keyword in the `target_link_library`, so they propagate downstream.

.gitlab-ci.yml
```diff
@@ -170,6 +170,7 @@ test:unit:
   script:
     - cmake -S . -B build ${CMAKE_OPTS}
     - cmake --build build ${CMAKE_BUILD_OPTS} --target run-unit-tests run-unit-tests-common
+    - cmake --build build ${CMAKE_BUILD_OPTS} --target run-python-wrapper-tests
```
stv0g, line 173:
```suggestion
    - cmake --build build ${CMAKE_BUILD_OPTS} --target run-unit-tests run-unit-tests-common run-python-wrapper-tests
```

clients/CMakeLists.txt
```diff
@@ -5,4 +5,5 @@
 # SPDX-License-Identifier: Apache-2.0
 
 add_subdirectory(opal-rt/rtlab-asyncip)
+add_subdirectory(python-wrapper)
```
stv0g, line 8: Usually such language support interfaces are called "bindings". Hence also the name "pybind".

I propose to rename this to "python-binding". Its not really wrapping anything in a sense that it builds on top or extends VILLASnode's functionality.

clients/python-wrapper/villas-python-wrapper.cpp
```diff
@@ -0,0 +1,246 @@
+/* Python-wrapper.
```
stv0g, line 1:
```suggestion
/* Python bindings.
```

clients/python-wrapper/villas-python-wrapper.cpp
```diff
@@ -0,0 +1,246 @@
...
+ * Author: Kevin Vu te Laar <vu.te@rwth-aachen.de>
+ * SPDX-FileCopyrightText: 2014-2025 Institute for Automation of Complex Power Systems, RWTH Aachen University
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <jansson.h>
+#include <pybind11/pybind11.h>
+#include <uuid/uuid.h>
+#include <villas/node.hpp>
+#include <villas/sample.hpp>
```
stv0g, line 12: Please separate headers in the following three sections:

- System headers
- Third-party library headers
- VILLASnode headers

```suggestion
#include <jansson.h>
#include <pybind11/pybind11.h>
#include <uuid/uuid.h>

#include <villas/node.hpp>
#include <villas/sample.hpp>
```

packaging/docker/Dockerfile.debian
```diff
@@ -51,7 +51,10 @@ RUN apt-get update && \
 		liblua5.3-dev \
 		libhiredis-dev \
 		libnice-dev \
-		libmodbus-dev
+		libmodbus-dev \
+		python3 \
```
stv0g, line 55: Depending on `python3` here is not required as its a direct dependency of `python3-dev`.
See: https://debian.pkgs.org/12/debian-main-arm64/python3-dev_3.11.2-1+b1_arm64.deb.html

E.g. we are also not depending on `libmodbus`, and only on `libmodbus-dev`.

packaging/docker/Dockerfile.debian-multiarch
```diff
@@ -58,7 +58,10 @@ RUN apt-get update && \
 		libusb-1.0-0-dev:${ARCH} \
 		liblua5.3-dev:${ARCH} \
 		libhiredis-dev:${ARCH} \
-		libmodbus-dev:${ARCH}
+		libmodbus-dev:${ARCH} \
+		python3:${ARCH} \
```
stv0g, line 62: Some as above.

packaging/docker/Dockerfile.ubuntu
```diff
@@ -59,7 +59,10 @@ RUN apt-get update && \
 		libmodbus-dev \
 		libre2-dev \
 		libglib2.0-dev \
-		libcriterion-dev
+		libcriterion-dev \
+		python3 \
```
stv0g, line 63: Same as above.

clients/python-wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+# Python-wrapper CMakeLists.
+#
+# Author: Kevin Vu te Laar <vu.te@rwth-aachen.de>
+# SPDX-FileCopyrightText: 2014-2025 Institute for Automation of Complex Power Systems, RWTH Aachen University
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.15...3.29)
```
stv0g, line 7: Could we move this to the top-level `CMakeLists.txt`? Or is there a reasoning for having this also here?

Building this as a dedicated project wont work due to the missing `villas` library target which you use below..

stv0g:
Hi @al3xa23 do you or your student need support in getting this merged?

al3xa23:
Hi, sorry, we need some more time here. If we need help, we come back to you. Thanks :)

.gitlab-ci.yml
```diff
@@ -173,7 +173,8 @@ test:unit:
   image: ${DOCKER_IMAGE_DEV}:${DOCKER_TAG}
   script:
     - cmake -S . -B build ${CMAKE_OPTS}
-    - cmake --build build ${CMAKE_BUILD_OPTS} --target run-unit-tests run-unit-tests-common
+    - export PYTHONPATH=$PYTHONPATH:${PWD}/build/clients/python-binding
+    - cmake --build build ${CMAKE_BUILD_OPTS} --target run-unit-tests run-unit-tests-common run-python-binding-tests
```
KeVteL, line 177: regarding:
https://github.com/VILLASframework/node/pull/884#discussion_r2033663485

I am not quite sure whether or not it is better to export the path of the binding `.so` as is done here or
to make use of something like this:
```yaml
paths:
  -build/clients/python-binding
  ```
  
  I am not sure whether or not this would even work.

**Comment by KeVteL** [comment]
regarding: https://github.com/VILLASframework/node/pull/884#discussion_r2033673569 https://github.com/VILLASframework/node/pull/884#discussion_r2033673878 https://github.com/VILLASframework/node/pull/884#discussion_r2033674228

I also removed the package `python` from `Dockerfile.fedora`, as `python-devel` also depends on it. b1e9a2f637f3430af9c9f2cb44ef3fdbbfdbac54

File: clients/python-wrapper/CMakeLists.txt (hunk 9 above)
Comment (KeVteL) [reply clients/python-wrapper/CMakeLists.txt]: addressed in: 6806d2409f5fba50dbed4b01589b20e518bf5787

File: packaging/docker/Dockerfile.rocky [hunk 11]
```diff
@@ -25,7 +25,8 @@ RUN dnf --allowerasing -y install \
 	flex bison \
 	texinfo git git-svn curl tar patchutils \
 	protobuf-compiler protobuf-c-compiler \
-	clang-tools-extra
+	clang-tools-extra \
+	python-devel
```
stv0g, line 29: Do you mind moving this below to the pybind11-devel package, as for the other Dockerfiles?

tests/unit/CMakeLists.txt
```diff
@@ -35,5 +35,13 @@ add_custom_target(run-unit-tests
     USES_TERMINAL
 )
 
+add_custom_target(run-python-binding-tests
```
stv0g, line 38: I am not so sure if this is really a unit-test, as its not testing individual and isolated parts of the code (e.g. functions).

Instead the Python test seems a bit more elaborate.

What do you think about adding it as an integration test? Or maybe as a dedicated test in itself?

I am interested in your point of view here :)

stv0g (changes requested):
Great work :) I have only minor points regarding packaging..

Please let me know if you need some support on this.

packaging/docker/Dockerfile.rocky
stv0g: Hi @KeVteL,

> I realized during a debian install, that the dependency list of the Dockerfiles debian, debian-multiarch and rocky are missing libwebsockets(-dev), which is required to build VILLASnode.

This maybe due to the version of libwebsocket shipped in the distributions package repository to be too old. If a package is not already installed (e.g. by the Dockerfile), we will install the missing dependencies (using the correct newer versions) via the `deps.sh` shell script.

However, I would be surprised if for the particular case of libwebsocket, the distro packages are too old. We should check here.

We also have upgrade some of the base distro versions (e.g. Debian 12 -> 13) in the last months without checking all the other dependencies. It might be that we now have the required versions already provided by the distros.

The exact version requirements are documented here:
- https://villas.fein-aachen.org/docs/node/installation#prerequisites
- https://github.com/VILLASframework/node/blob/9b81d28fe1ce03c40393251837b524087c50b03c/CMakeLists.txt#L95

al3xa23:
Hi @stv0g, can you please give some (final?) comment here so that we can proceed? Thanks :)

al3xa23:
Hi @stv0g, what do you think here? Can we proceed? What is needed from your side? 
Let us know, thanks :)

clients/python-binding/CMakeLists.txt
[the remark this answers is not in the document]
leonardocarreras: Hi, I know it is a year later... But some extra comments here... In the end my understanding is that you need to use the PyPI tools to make your life easier
- For the case of how to do this with the tools, I would like to point to [cibuildwheel](https://cibuildwheel.pypa.io/en/stable/)
    - If you need an example of this, you can check DPsim [here](https://github.com/sogno-platform/dpsim/blob/b63e7153f2cc94ad8844e1110efeb7e2ac9c13e9/.github/workflows/publish_to_pypi.yaml#L41)
    - You need to provide your own Containerfile like [here](https://github.com/sogno-platform/dpsim/blob/master/packaging/Docker/Dockerfile.manylinux)
- About the CMake compilation, this mirrors a bit the way villas does, so you need the villas libraries and the python code to get it packaged, plus the tooling will take care to include the libraries inside the package
    - About the code of the repo and how to compile, that is not a problem once you use the tooling and you change the way that is done, you can use a [pyproject.toml](https://github.com/sogno-platform/dpsim/blob/master/pyproject.toml) that runs a [setup.py](https://github.com/sogno-platform/dpsim/blob/master/setup.py) which does the CMake compilation, and select the respective targets that will be part of the package
    - You need to make sure that libraries are available in the container, and that you include the python code like [here](https://github.com/sogno-platform/dpsim/blob/b63e7153f2cc94ad8844e1110efeb7e2ac9c13e9/pyproject.toml#L58)

python/binding/capi_python_binding.cpp
```diff
@@ -0,0 +1,404 @@
+/* Python-binding.
+ *
+ * Author: Kevin Vu te Laar <vu.te@rwth-aachen.de>
+ * SPDX-FileCopyrightText: 2014-2025 Institute for Automation of Complex Power
```
leonardocarreras, line 4: 2025? I think this file is new

python/binding/capi_python_binding.cpp
```diff
@@ -0,0 +1,404 @@
...
+}
+
+/* pybind11 can not deal with (void **) as function input parameters,
+ * therefore cast a simple (void *) pointer to the corresponding type
+ *
+ * wrapper bindings, sorted alphabetically
+ * @param villas_node   Name of the module to be bound
+ * @param m             Access variable for modifying the module code
+ */
+PYBIND11_MODULE(python_binding, m) {
```
leonardocarreras, line 127: any specific reason why all is prefixed with `node_`

python/binding/capi_python_binding.cpp
```diff
@@ -0,0 +1,404 @@
...
+  unsigned int size() const { return len; }
+
+private:
+  vsample **smps;
+  unsigned int len;
+};
+
+struct timespec ns_to_timespec(int64_t time_ns) {
+  struct timespec ts;
+  ts.tv_nsec = time_ns / 1'000'000'000LL;
```
leonardocarreras, line 115: this looks flipped, isnt it nsec the reminder, and sec the division?

ts.tv_nsec = time_ns / 1'000'000'000LL;
ts.tv_sec  = time_ns % 1'000'000'000LL;

python/binding/capi_python_binding.cpp
```diff
@@ -0,0 +1,404 @@
...
+    this->len = new_len;
+
+    return new_len;
+  }
+
+  int clear() {
+    if (this->smps) {
+      unsigned int i = 0;
+      for (; i < len; ++i) {
+        sample_decref(smps[i]);
```
leonardocarreras, line 86: no check for` smps[i]` is null

python/binding/capi_python_binding.cpp
```diff
@@ -0,0 +1,404 @@
...
+    d["data"] = data;
+
+    return d;
+  });
+
+  py::class_<SamplesArray>(m, "SamplesArray")
+      .def(py::init<unsigned int>(), py::arg("len"))
+      .def("__getitem__",
+           [](SamplesArray &a, unsigned int idx) {
+             assert(idx < a.size() && "Index out of bounds");
```
leonardocarreras, line 388: better throw an exception

python/binding/capi_python_binding.cpp
```diff
@@ -0,0 +1,404 @@
...
+             return a[idx];
+           })
+      .def("__setitem__",
+           [](SamplesArray &a, unsigned int idx, void *smp) {
+             assert(idx < a.size() && "Index out of bounds");
```
leonardocarreras, line 393: better throw an exception

python/binding/capi_python_binding.cpp
```diff
@@ -0,0 +1,404 @@
...
+      "sample_pack",
+      [](const py::list values, std::optional<int64_t> ts_origin_ns,
+         std::optional<int64_t> ts_received_ns, unsigned seq = 0) -> void * {
+        struct timespec ts_origin =
+            ts_origin_ns ? ns_to_timespec(*ts_origin_ns) : time_now();
+        struct timespec ts_received =
+            ts_received_ns ? ns_to_timespec(*ts_received_ns) : time_now();
+
+        unsigned values_len = values.size();
+        double cvalues[values_len];
```
leonardocarreras, line 312: can we use std::vector here?

lib/node_capi.cpp
```diff
@@ -169,7 +169,7 @@ vsample *sample_pack(unsigned seq, struct timespec *ts_origin,
   return (vsample *)smp;
 }
 
-void sample_unpack(vsample *s, unsigned *seq, struct timespec *ts_origin,
+void sample_unpack(vsample *s, uint64_t *seq, struct timespec *ts_origin,
```
leonardocarreras, line 172: this breaks the C API, is this needed?

lib/node_capi.cpp
```diff
@@ -152,12 +152,12 @@ unsigned sample_length(vsample *s) {
   return smp->length;
 }
 
-vsample *sample_pack(unsigned seq, struct timespec *ts_origin,
+vsample *sample_pack(uint64_t *seq, struct timespec *ts_origin,
```
leonardocarreras, line 155: this breaks the C API, is this needed?

include/villas/node.h
```diff
@@ -67,11 +68,11 @@ unsigned sample_length(vsample *smp);
 
 void sample_decref(vsample *smp);
 
-vsample *sample_pack(unsigned seq, struct timespec *ts_origin,
+vsample *sample_pack(uint64_t *seq, struct timespec *ts_origin,
```
leonardocarreras, line 71: this breaks the C API, is this needed?

include/villas/node.h
```diff
@@ -67,11 +68,11 @@ unsigned sample_length(vsample *smp);
...
                      struct timespec *ts_received, unsigned len,
                      double *values);
 
-void sample_unpack(vsample *s, unsigned *seq, struct timespec *ts_origin,
+void sample_unpack(vsample *s, uint64_t *seq, struct timespec *ts_origin,
```
leonardocarreras, line 75: this breaks the C API, is this needed?

include/villas/node.h
```diff
@@ -6,6 +6,7 @@
  */
 
 #pragma once
+#include <cstdint>
```
leonardocarreras, line 9: can we use `stdint.h `?

python/binding/capi_python_binding.cpp
leonardocarreras, line 4: do not break this line

python/binding/capi_python_binding.cpp
```diff
@@ -0,0 +1,404 @@
+/* Python-binding.
+ *
+ * Author: Kevin Vu te Laar <vu.te@rwth-aachen.de>
+ * SPDX-FileCopyrightText: 2014-2025 Institute for Automation of Complex Power
+ * Systems, RWTH Aachen University SPDX-License-Identifier: Apache-2.0
```
leonardocarreras, line 5: the SPDX-License should be in a new line

tests/integration/CMakeLists.txt
```diff
@@ -18,4 +18,12 @@ add_custom_target(run-integration-tests
...
-add_dependencies(run-tests run-integration-tests)
+add_custom_target(run-python-integration-tests
+    COMMAND
+        python3 -m unittest discover ${CMAKE_CURRENT_SOURCE_DIR}
+    DEPENDS
+        python_binding
+    USES_TERMINAL
+)
+
+add_dependencies(run-tests run-integration-tests run-python-integration-tests)
```
leonardocarreras, line 29: missing python_bindings dependency? will fail if that is not there

.gitlab-ci.yml
```diff
@@ -182,6 +182,28 @@ test:cppcheck:
       - cppcheck.log
     expose_as: cppcheck
 
+test:python_unit_integration:
```
leonardocarreras, line 185: this need to be converted to cibuildwheels and github action

tests/unit/CMakeLists.txt
```diff
@@ -35,5 +35,13 @@ add_custom_target(run-unit-tests
     USES_TERMINAL
 )
 
+add_custom_target(run-python-unit-tests
+    COMMAND
+        python3 -m unittest discover ${CMAKE_CURRENT_SOURCE_DIR}
+    DEPENDS
```
leonardocarreras, line 41: `DEPENDS` does not work as `add_dependencies`, is that what you meant?

tests/unit/python/test_python_binding.py
```diff
@@ -0,0 +1,239 @@
+"""
+Author: Kevin Vu te Laar <vu.te@rwth-aachen.de>
+SPDX-FileCopyrightText: 2014-2025 Institute for Automation of Complex Power Systems, RWTH Aachen University
```
leonardocarreras, line 3: 2025?

tests/unit/python/test_binding_wrapper.py
```diff
@@ -0,0 +1,241 @@
+"""
+Author: Kevin Vu te Laar <vu.te@rwth-aachen.de>
+SPDX-FileCopyrightText: 2014-2025 Institute for Automation of Complex Power Systems, RWTH Aachen University
```
leonardocarreras, line 3: 2025?

tests/unit/python/test_binding_wrapper.py
```diff
@@ -0,0 +1,241 @@
...
+SPDX-License-Identifier: Apache-2.0
+"""  # noqa: E501
```
leonardocarreras, line 5: what is noqa: E501

tests/integration/CMakeLists.txt
```diff
@@ -18,4 +18,12 @@ add_custom_target(run-integration-tests
         villas-hook
 )
 
-add_dependencies(run-tests run-integration-tests)
+add_custom_target(run-python-integration-tests
+    COMMAND
+        python3 -m unittest discover ${CMAKE_CURRENT_SOURCE_DIR}
+    DEPENDS
```
leonardocarreras, line 24: DEPENDS vs add_dependencies

tests/integration/python/test_python_binding.py
```diff
@@ -0,0 +1,287 @@
+"""
+Author: Kevin Vu te Laar <vu.te@rwth-aachen.de>
+SPDX-FileCopyrightText: 2014-2025 Institute for Automation of Complex Power Systems, RWTH Aachen University
```
leonardocarreras, line 3: 2025?

tests/integration/python/test_binding_wrapper.py
```diff
@@ -0,0 +1,305 @@
...
+            self.assertEqual(len(node), 100)
+            node[199].pack_from([1.01, 2.01, 3.01, 4.01, 5.01], int(1e9), int(1e9) + 100)
+            self.assertEqual(len(node), 200)
+            node[199].unpack_to(node[299])
+            self.assertEqual(len(node), 300)
+        except Exception as e:
+            self.fail(f"err: {e}")
+
+
+test_node_config = {
```
leonardocarreras, line 211: can we store the configs somewhere? they also seem to be quite similar one to each other

CMakeLists.txt
leonardocarreras, line 216: Add a WITH_PYTHON_BINDINGS option

.gitlab-ci.yml
```diff
@@ -164,7 +164,7 @@ test:python:
     - pytest --verbose .
     - black --line-length=90 --extend-exclude=".*(\\.pyi|_pb2.py)$" --check .
     - flake8 --max-line-length=90 --extend-exclude="*.pyi,*_pb2.py" .
-    - mypy .
+    - mypy --explicit-package-bases .
```
leonardocarreras, line 167: pybind provides a way to generate stubs, would we want to do this still with mypy?

leonardocarreras (commented):
Hi @KeVteL! Thanks for your PR, we just took a new impulse and want to help to get our contributions merged... So sorry already for the delays...

I have some comments, many are questions, many are preferences... Is not meant to be something definitive or a disagreement, but a starting point to think about the direction...
- I have the feeling that we can use CMakeExtension or scikit-build helps with getting built all together
- If the previous is true, you can use cibuildtools so then this is easier
- I prefer pybind11-stubgen to hand-written pyi, I think even now some of the stubs are out of sync with their functions/classes
- Using auditwheel can help patch the package and include the libraries, so that is also not a problem theoretically
- What exactly to bind is still to me a question but I would say is more on the direction of villas/node.hpp, I could be missing something... Many new functions and classes somehow reimplement the logic of the c/c++ code but in python... Maybe this is something to discuss more... But I have the feeling that we could avoid this layering that duplicates, ref counters and pack/unpack...
- Node is somehow declared in many places as of the state now, we could end up shadowing one with each other unadvertedly in my opinion
- Some more nitpicky things like the years and lines on SPDX (new files could have only 2025 I guess, unless something different is asked in the contributing guidelines)
- Maybe configs can be separated from their tests when they are too similar, to avoid duplication


I think that it would be nice to know your opinions and takes about those items. Thanks in advance!

leonardocarreras:
In case it helps, I started a discussion here #1029 

Happy to know your opinions @al3xa23 @stv0g @KeVteL

PR #888, opened 2025-04-16
Fixes to compile in Fedora 42

leonardocarreras:
This introduces changes to comply with some stricter rules on the compiler.
- The template definiton was moved before the hack in the register_types
- Override instead of hide for the dump() function


leonardocarreras:
@stv0g, if you have time for a review, will appreciate

stv0g (approved):

stv0g:
Thanks :)

PR #887, opened 2025-04-09
feat: Add OpenDSS node type

JitpanuM:
TODO:

- Add integration test
- Add OpenAPI spec
- Add Documentation

PR #889, opened 2025-04-17
Add new node-type for OpenDSS

stv0g:
Continuation of #887 

TODO:

Add integration test
Add OpenAPI spec
Add Documentation


stv0g:
@JitpanuM I think, I got most of the build issues fixed. But there is a problem with the example config and the integration test thereof. Could you have a look?

JitpanuM:
It seem like opendss change the working directory to /home/user/OpenDSS. That is why the integration test couldn't find the config file.

stv0g:
Ouch, that's annoying. Do you already have an idea how we can fix it?

stv0g:
Urgh, we get some flaky / broken integration tests now :(

I suspect that something with the timing changed.. Maybe loading the `libopendssc.so` library or any of the other version updates is causing this :(

We will need to split-up this PR into smaller ones unfortunately :(

stv0g:
I moved a bunch of unreleated updates to this PR #893

stv0g:
Okay, the Docker image updates (#893) have been merged into master, and I have rebased this PR onto it.

I also did some cleanup of the Git history. Please reset your local branch before pushing any more changes :) Thanks :)

stv0g:
Okay, two more fixes:

- For some reason a member variable of an unrelated hook was not properly initialized and caused a test to fail
- For some other reason OpenDSS is seeding the random number generator differently than VILLASnode did in the past. This broke another test.

PR #890, opened 2025-04-24
Add Prometheus exporter

stv0g:


stv0g:
Hi @SystemsPurge,

while adding an integration test for your new API endpoint, I noticed that it only exposes a fraction of the available metrics.

Would you be open to extending it to all available metrics?

E.g., here is the code which prints the periodic stats to the console:
https://github.com/VILLASframework/node/blob/cc52c765425b54eebdddd6d545472e3323c94456/lib/stats.cpp#L145

But this is still only a fraction, as most of the statistics are backed by a histogram:
https://github.com/VILLASframework/node/blob/master/common/lib/hist.cpp

Prometheus also has quite good support for histograms:
https://prometheus.io/docs/practices/histograms/

I think it would be pretty neat, if we could export the full histograms to Prometheus :)
That way we could render a nice RTT/One-way-delay distribution in Grafana :)

Here is an example which plots the histogram data even over time:

image

SystemsPurge:
@stv0g 
I was hesitant about exporting histograms because it seemed a tiny bit innefficient, since you say it's ok i will do it.

stv0g:
Performance-wise I think it should be fine..

We might have to check if we can directly use the data from the histograms, because they are collecting their counts over the whole runtime, rather than a moving windows (e.g. last 5minutes).

We probably need to check, but I believe Prometheus expects the counts/bucket over a certain time frame / moving window..

SystemsPurge:
@stv0g  With histograms specifically it does not seem to require a timestamp
However one possible piece of missing information is sum.
If my understanding is correct the histograms does not hold values associated with the buckets. One therefore only gets metrics in percentile form in this case.
It should be done by the way, you can take a look

PR #891, opened 2025-04-24
Cleanups

stv0g:
This PR reformats the VILLASnode code with clang-format as well as adds missing executable bits on some of the scripts in `tools/`.


stv0g:
Included in #892

PR #894, opened 2025-04-27
Add support for newer cmake versions (and their policies)

leonardocarreras:
This PR adds:
- A check for CUDA language availability in CMakeLists of the main villlas-node
- Changes the hardcoding of a path in for villas-fpga (in gpu)

This gives some errors related to CMP0146 and CMP0104 in newer CMake versions

This is needed in sogno-platform/dpsim#334 (was tested locally)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,12 +6,25 @@
 
 cmake_minimum_required(VERSION 3.14)
 
-project(villas-node
-    VERSION 0.0.0
-    DESCRIPTION "VILLASnode"
-    HOMEPAGE_URL "https://www.fein-aachen.org/projects/villas-node/"
-    LANGUAGES C CXX
-)
+include(CheckLanguage)
+# Check if CUDA is available
+check_language(CUDA)
+
+if(CMAKE_CUDA_COMPILER)
+    project(villas-node
+        VERSION 0.0.0
+        DESCRIPTION "VILLASnode"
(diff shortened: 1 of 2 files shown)


leonardocarreras:
The CI is not fully working (does not pull the images for the tests, for example for mosquitto)

stv0g (approved):

stv0g:
Hey @n-eiling,

do you want to review this change as well? Are you still working with VILLASnode? Or do you want to be removed from the CODEOWNERS file for future PRs?

stv0g:
Hey @leonardocarreras,

lets wait until the end of the week and otherwise merge it :)

n-eiling:
I don't think the GPU code is working right now. As far as I know, nobody has used it in a long time. I wouldn't be opposed to removing it.

n-eiling (approved):

leonardocarreras:
@n-eiling thanks for the feedback, it is still giving some errors even with this... If removing is also a reasonable option then is a good future step, `gdrcopy` also brings problems

stv0g:
Do we then rip-out all the GPU related stuff?

n-eiling:
It's not that deeply integrates, or is it? If it's not too much trouble, I think removing it is a good idea.

leonardocarreras:
I can attempt a PR to remove the GPU stuff this week. I assume if it works, in that case this PR will not be needed...

leonardocarreras:
Closing now, not needed due to #908

PR #897, opened 2025-05-05
Add support for R-GOOSE messages

pjungkamp:
# TODO

- [x] R-GOOSE Receiver
- [x] R-GOOSE Publisher


stv0g (changes requested):

include/villas/nodes/iec61850_goose.hpp
```diff
@@ -27,9 +28,7 @@
 #include <villas/queue_signalled.h>
 #include <villas/signal.hpp>
 
-namespace villas {
-namespace node {
-namespace iec61850 {
+namespace villas::node::iec61850 {
```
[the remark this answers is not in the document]
pjungkamp: Would you be open to doing some refactoring using `clang-tidy`? It can do the project-wide collapsing of namespaces automatically! https://clang.llvm.org/extra/clang-tidy/checks/modernize/concat-nested-namespaces.html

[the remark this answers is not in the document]
stv0g: Yes for sure 👍 Every tool-assisted refactoring is a big win.

Can we do this in a separate PR, which also checks the adherence to this convention in Future commits in the CI?

pjungkamp:
@stv0g Would you be ok with the `libiec61850` and `lib60870` fixups for `mbedtls` support being in the `default` overlay until I manage to get these into `nixpkgs`?

packaging/deps.sh
```diff
@@ -194,6 +194,12 @@ fi
 if ! pkg-config "libiec61850 >= 1.6.0" && \
     should_build "iec61850" "for the iec61850 node-type"; then
     git clone ${GIT_OPTS} --branch v1.6.0 https://github.com/mz-automation/libiec61850.git
+
+    pushd libiec61850/third_party/mbedtls/
+    wget https://github.com/Mbed-TLS/mbedtls/archive/refs/tags/v3.6.0.tar.gz
```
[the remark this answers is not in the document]
stv0g: I believe we were using `wget` as some platforms shipped it instead of curl by default (I believe its part of busybox).

But I am fine with switching to curl instead. As we are nowadays using pre-compiled Nix builds on those platforms (OPAL-RT Linux).

stv0g:
> @stv0g Would you be ok with the `libiec61850` and `lib60870` fixups for `mbedtls` support being in the `default` overlay until I manage to get these into `nixpkgs`?

Yes please  🙏🏻  I would favor velocity in this case, instead of beeing blocked by Nixpkgs maintainers.

pjungkamp:
@stv0g I consider this PR ready. Could you do another review? I think the failing [pkg:docker: [debian, linux/arm64/v8, arm64, aarch64-linux-gnu, pi5]](https://git.rwth-aachen.de/acs/public/villas/node/-/jobs/6500761) pipeline is not caused by this PR but rather by some broken cached state in the debian image on the Pi5.

tests/integration/node-rgoose.sh
stv0g, line 1: Can we prefix this test with the name of the node-type?

E.g. `node-iec61850-8-1-rgoose`?

tests/integration/node-rgoose.sh
```diff
@@ -0,0 +1,141 @@
...
+  ]
+}
+EOF
+
+villas signal -l ${NUM_SAMPLES} -n counter -r 10 > input.dat
+
+VILLAS_LOG_PREFIX="[node] " \
+villas node config.json &
+
+cat input.dat >&2
```
stv0g, line 129: Do we need this? I think, we omitted printing test data in other tests too.

tests/integration/node-rgoose.sh
```diff
@@ -0,0 +1,141 @@
...
+
+# Wait for node to complete
+sleep $((NUM_SAMPLES / 10 + 2))
+
+kill %%
+wait %%
+
+cat output.dat >&2
```
stv0g, line 137: Do we need this? I think, we omitted printing test data in other tests too.

flake.nix
```diff
@@ -110,11 +110,30 @@
 
       # Standard flake attribute allowing you to add the villas packages to your nixpkgs
       overlays = {
-        default = final: prev: packagesWith final;
+        default = final: prev: (packagesWith final) // {
+          libiec61850 = prev.libiec61850.overrideAttrs {
```
stv0g, line 114: It would be nice if we could move this overlay into separate `.nix` file somewhere under `packaging/nix`. Ideally with a short comment on what it does.

stv0g: Thanks. I think the current state is good enough :)

PR #910, opened 2025-05-10
Fixes OpenDSSC installation: correct find condition and add to linker path

leonardocarreras:
The OpenDSCC installs in a non-usual path. 

The current check with find will always install because is not looking into the path where it gets installed.

The symbols needed are not found by the linker, so this is added.

image

### DIFF
--- a/packaging/deps.sh
+++ b/packaging/deps.sh
@@ -484,7 +484,7 @@ if ! pkg-config "libmodbus >= 3.1.0" && \
     popd
 fi
 
-if ! find /usr/{local/,}{lib,bin} -name "libOpenDSSC.so" | grep -q . &&
+if ! find /usr/local/ -name "libOpenDSSC.so" | grep -q . &&
     should_build "opendss" "For opendss node-type"; then
     git svn clone -r 4020:4020 https://svn.code.sf.net/p/electricdss/code/trunk/VersionC OpenDSS-C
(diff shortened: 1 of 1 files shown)


stv0g:
Hi @leonardocarreras,

I think this should already have been resolved by this change which I merged a few days ago:
https://github.com/VILLASframework/node/commit/8bc1fe2a1fa7b5baa94e7ce5f9f16ce0032370ed#diff-49cac9549226aeff7ab6461609d9afc6d8bec062d76414c79007b101d5c2eda6R68

I am not a big fan of modifying the linker search path in `deps.sh`, since this is mixing up concerns:
https://github.com/VILLASframework/node/pull/910/files#diff-3702afcced7ceb56a7ccf7ebb5f0cfd309d51c83ba2cc77786507b02635eb291R506

A better solution, would be to add some more patches to install the library in the correct location. Like we already do here with other patches for OpenDSSC:
https://github.com/VILLASframework/node/blob/366df989cd2dfa1ffadad6c9076cf49595c320bc/packaging/deps.sh#L492

leonardocarreras:
Hi @stv0g! I understand your point, what happens is that when just installing villas dependencies with the script in your own machine or in a container, you do not realize that there is a problem until later (in my experience villas compiles, but does have a runtime problem)... I will then keep doing the same in the container for dpsim that uses villas...

PR #911, opened 2025-05-13
Enforce project wide `#include` order

pjungkamp:
This instructs `clang-format` to enforce a uniform `#include` order across all files. See https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includecategories

1. `^<villas/`: headers
2. `^<[[:lower:]]+>$`: standard library headers
3. `^<.*>$`: other system headers 
4. `^".*"$`: local headers

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -7,3 +7,18 @@ BasedOnStyle: LLVM
 # Disable automatic line-breaks in comments
 # as this breaks SPDX headers
 ReflowComments: false
+
+# Regroup #include directives to villas - std - other
+IncludeBlocks: Regroup
+IncludeIsMainRegex: / # disable main header heuristic
+IncludeCategories:
+  - Regex: '^<villas/'
+    Priority: 3
+    CaseSensitive: true
+  - Regex: '^<[[:lower:]_]+>$'
+    Priority: 1
+    CaseSensitive: true
+  - Regex: '^<.*>$'
+    Priority: 2
+  - Regex: '^".*"$'
+    Priority: 4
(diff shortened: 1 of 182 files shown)


n-eiling:
Didn't we want to keep the clang format file as simple and as standard as possible? Is this really worth it?

pjungkamp:
I've been asked to reorder includes in multiple PRs, about half the files in this repository contain some form of error regarding our current `std - other - villas` header order.

I just want `clang-format` to sort the includes for me instead of getting a "please reorder includes" review in every second PR.

The enforced order is of cause up for debate, if you want to stick with `std - other - villas`.

n-eiling:
Who asks you to reorder? You should not need to do anything style related that is not mentioned here: https://villas.fein-aachen.org/docs/node/development/contributing/. This mentions nothing about include order.
From the code-perspective include order shouldn't matter?

stv0g:
I think we agreed in the past to follow the LLVM/Clang coding standards.

The minimal clang-format configuraiton which we had in the past is not sufficient, as it can not identify our local headers (e.g. `#include <villas/...`).

So, believe extending the clang-format config is justified as long as we stick with our goal of following the Clang/LLVM coding standard.

pjungkamp:
The LLVM `#include` style can be found here: https://llvm.org/docs/CodingStandards.html#include-style

I deviate from these on the "main" header of a source file. `clang-format` has built-in heuristics to determine a source file's "main" header. But these work best when project includes use quotes instead of angle brackets for local headers.

I've found that these heuristics determined e.g. `<sys/socket.h>` as the "main" include for `clients/opal-rt/rtlab-asyncip/models/send_receive/src/socket.c`, so I didn't enable it in this PR.

PR #915, opened 2025-05-15
Binding test formatted with black(-jupyter)

KeVteL:
Manually formatted the Python binding test with black. The pipeline formatting resulted in an invalid escape sequence `\m`.

Oddly enough there is no such character sequence within the code. Especially not at the reported line 108 when looking at:
https://git.rwth-aachen.de/acs/public/villas/node/-/jobs/6248036
https://git.rwth-aachen.de/acs/public/villas/node/-/jobs/6249567

`black-jupyter` seems to be stricter than plain `black` since even formatting with plain `black` causes the pipeline to fail, as `black-jupyter` may still change the formatting of the file.

### DIFF
--- a/tests/unit/python_binding.py
+++ b/tests/unit/python_binding.py
@@ -10,8 +10,8 @@ import unittest
 import uuid
 import villas_node as vn
 
-class SimpleWrapperTests(unittest.TestCase):
 
+class SimpleWrapperTests(unittest.TestCase):
     def setUp(self):
         try:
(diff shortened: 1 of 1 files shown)


stv0g (approved):

stv0g:
@KeVteL I gave you direct access to the VILLASnode repo. This should allow you to directly push to the feature branch and trigger the pipeline in your own.

For merging the PRs, the usual PR approval and rules still apply.

PR #917, opened 2025-05-20
JOSS

al3xa23:
I would like to merge joss branch with master. 

If you prefer otherwise, it is also a possible to leave it in an extra branch as described here https://joss.readthedocs.io/en/latest/submitting.html#submission-requirements.


n-eiling:
Thanks again for all the effort. I think I'd prefer not putting the paper in the root of the repo. We have documentation and code separated, and I view this more as documentation.

stv0g:
I also agree with @n-eiling here. Does JOSS expect a specific path for the files? Or do they allow other locations as well?

n-eiling:
Are you okay with creating a `joss` branch and putting the .md there? Maybe we can additionally find a good place in the documentation repo?

al3xa23:
Sure, everything can stay right now in the JOSS branch. Then, I will close the PR. Thanks for the answers :)

PR #919, opened 2025-05-26
fix: Integration test for mqtt in docker container

RiteshKarki27:


stv0g:
We are managing the start-up of auxiliary services via the GitLab CI config here:
https://github.com/VILLASframework/node/blob/9b81d28fe1ce03c40393251837b524087c50b03c/.gitlab-ci.yml#L193

I dont believe this change here is necessary.

For local development setups, I recommend that you install the broker via the usual service managers on your system (e.g. SystemD)

PR #922, opened 2025-06-11
Option to read file at once. 

RiteshKarki27:
- Add the option to read entire file at once in file_start. 
- This does not affect the way the file is handled during real-time operation.  
- Include an option in the configuration ("read_mode") to activate this operation. 

@al3xa23

### DIFF
--- a/lib/nodes/file.cpp
+++ b/lib/nodes/file.cpp
@@ -5,10 +5,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include "villas/sample.hpp"
 #include <cerrno>
+#include <cstdio>
 #include <cstring>
-
+#include <iostream>
 #include <libgen.h>
+#include <ostream>
+#include <string>
 #include <sys/stat.h>
 #include <unistd.h>
 
@@ -76,16 +80,17 @@ int villas::node::file_parse(NodeCompat *n, json_t *json) {
   const char *uri_tmpl = nullptr;
   const char *eof = nullptr;
   const char *epoch = nullptr;
+  const char *read_mode = nullptr;
   double epoch_flt = 0;
 
   ret = json_unpack_ex(json, &err, 0,
                        "{ s: s, s?: o, s?: { s?: s, s?: F, s?: s, s?: F, s?: "
-                       "i, s?: i }, s?: { s?: b, s?: i } }",
+                       "i, s?: i, s?: s }, s?: { s?: b, s?: i } }",
                        "uri", &uri_tmpl, "format", &json_format, "in", "eof",
                        &eof, "rate", &f->rate, "epoch_mode", &epoch, "epoch",
                        &epoch_flt, "buffer_size", &f->buffer_size_in, "skip",
                        &f->skip_lines, "out", "flush", &f->flush, "buffer_size",
-                       &f->buffer_size_out);
+                       &f->buffer_size_out, "read_mode", &read_mode);
   if (ret)
     throw ConfigError(json, err, "node-config-node-file");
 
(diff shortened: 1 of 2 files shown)


lib/nodes/file.cpp
```diff
@@ -5,10 +5,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include "villas/sample.hpp"
```
stv0g, line 8:
```suggestion
#include <villas/sample.hpp>
```

stv0g: Please also make sure includes are properly ordnered.

lib/nodes/file.cpp
```diff
@@ -127,6 +132,15 @@ int villas::node::file_parse(NodeCompat *n, json_t *json) {
       throw RuntimeError("Invalid value '{}' for setting 'epoch'", epoch);
   }
 
+  if (read_mode) {
+    if (!strcmp(read_mode, "all")) {
+      n->logger->debug("read mode: all set");
+      f->read_mode = file::ReadMode::READ_ALL;
+    } else {
+      f->read_mode = file::ReadMode::RATE_BASED;
+      n->logger->debug("read mode: rate based set");
```
stv0g, line 141:
```suggestion
```

lib/nodes/file.cpp
```diff
@@ -127,6 +132,15 @@ int villas::node::file_parse(NodeCompat *n, json_t *json) {
       throw RuntimeError("Invalid value '{}' for setting 'epoch'", epoch);
   }
 
+  if (read_mode) {
+    if (!strcmp(read_mode, "all")) {
+      n->logger->debug("read mode: all set");
+      f->read_mode = file::ReadMode::READ_ALL;
```
stv0g, line 138:
```suggestion
```

stv0g: Let put this into `file_print()` (`Node::getDetails()`).

Here: https://github.com/VILLASframework/node/blob/aaece4c1b8e27ca883b1e5d0b55e6721ada6d2b8/lib/nodes/file.cpp#L147

lib/nodes/file.cpp
```diff
@@ -298,12 +312,43 @@ int villas::node::file_start(NodeCompat *n) {
 
   sample_free(smp);
 
+  if (f->read_mode == file::ReadMode::READ_ALL) {
+    n->logger->info("Reading entire file into buffer");
+
+    struct Sample *smp_buffer;
```
stv0g, line 318:
```suggestion
    Sample *smp_buffer;
```

stv0g (changes requested):
This is a great improvement :) We definitetly need this to improve VILLASnode's real-time performance!

I did a review of the code but noticed that your changes do not play together with the following options of the file node-type:

- `eof_mode`
- `skip_lines`
- `epoch`
- `epoch_mode`

For this to be merged, I would like to have `read_mode` work as expected with the other options of the file-node type.

lib/nodes/file.cpp
```diff
@@ -320,6 +359,62 @@ int villas::node::file_read(NodeCompat *n, struct Sample *const smps[],
 
   assert(cnt == 1);
 
+  if (f->read_mode == file::ReadMode::READ_ALL) {
```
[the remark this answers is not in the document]
RiteshKarki27: Yes, I thought so too. I have been testing with lambda functions and I can replace the duplicated code with a lambda function instead if that is allowed.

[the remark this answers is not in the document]
RiteshKarki27: Yeah, sure. :)

PR #923, opened 2025-06-12
Add new OPAL-RT Orchestra node-type

stv0g:
This PR is adding a new node-type for improving the interface between VILLASnode and OPAL-RT software platforms using OPAL-RT Orchestra Co-simulation framework.

We are using this new node-type both in the ENSURE and SEGuRo projects.

See:
- https://opal-rt.atlassian.net/wiki/spaces/PRD/pages/505348099/Orchestra+I+O
- https://opal-rt.atlassian.net/wiki/spaces/PRD/pages/143626248/Orchestra
- https://opal-rt.atlassian.net/wiki/spaces/PRD/pages/143787939/RT-LAB+API+Orchestra

### DIFF
--- a/packaging/nix/villas.nix
+++ b/packaging/nix/villas.nix
@@ -29,7 +29,7 @@
   withNodeMqtt ? withAllNodes,
   withNodeNanomsg ? withAllNodes,
   withNodeOpenDSS ? withAllNodes,
-  withNodeOpalOrchestra ? withAllNodes,
+  withNodeOpalOrchestra ? (withAllNodes && system == "x86_64-linux"),
   withNodeRedis ? withAllNodes,
   withNodeRtp ? withAllNodes,
   withNodeSocket ? withAllNodes,
@@ -145,59 +145,60 @@ stdenv.mkDerivation {
     protobufcBuildBuild
   ];
 
-  buildInputs =
-    [
-      libwebsockets
-      openssl
-      curl
-      spdlog
-      bash
-    ]
-    ++ lib.optionals withExtraTesting [
-      boxfort
-      criterion
-      libffi
-      libgit2
-      pcre2
-    ]
-    ++ lib.optionals withExtraGraphviz [ graphviz ]
-    ++ lib.optionals withHookLua [ lua ]
-    ++ lib.optionals withNodeAmqp [ rabbitmq-c ]
-    ++ lib.optionals withNodeComedi [ comedilib ]
-    ++ lib.optionals withNodeEthercat [ ethercat ]
-    ++ lib.optionals withNodeIec60870 [ lib60870 ]
-    ++ lib.optionals withNodeIec61850 [ libiec61850 ]
-    ++ lib.optionals withNodeKafka [
-      rdkafka
-      cyrus_sasl
-    ]
-    ++ lib.optionals withNodeModbus [ libmodbus ]
-    ++ lib.optionals withNodeMqtt [ mosquitto ]
-    ++ lib.optionals withNodeNanomsg [ nanomsg ]
-    ++ lib.optionals withNodeOpenDSS [ opendssc ]
-    ++ lib.optionals withNodeOpalOrchestra [ orchestra libxml2 ]
-    ++ lib.optionals withNodeRedis [ redis-plus-plus ]
-    ++ lib.optionals withNodeRtp [ libre ]
-    ++ lib.optionals withNodeSocket [ libnl ]
-    ++ lib.optionals withNodeTemper [ libusb1 ]
-    ++ lib.optionals withNodeUldaq [ libuldaq ]
-    ++ lib.optionals withNodeWebrtc [ libdatachannel ]
-    ++ lib.optionals withNodeZeromq [
-      czmq
-      libsodium
-    ];
+  buildInputs = [
+    libwebsockets
+    openssl
+    curl
+    spdlog
+    bash
+  ]
+  ++ lib.optionals withExtraTesting [
+    boxfort
+    criterion
+    libffi
+    libgit2
+    pcre2
+  ]
+  ++ lib.optionals withExtraGraphviz [ graphviz ]
+  ++ lib.optionals withHookLua [ lua ]
+  ++ lib.optionals withNodeAmqp [ rabbitmq-c ]
+  ++ lib.optionals withNodeComedi [ comedilib ]
+  ++ lib.optionals withNodeEthercat [ ethercat ]
+  ++ lib.optionals withNodeIec60870 [ lib60870 ]
+  ++ lib.optionals withNodeIec61850 [ libiec61850 ]
+  ++ lib.optionals withNodeKafka [
+    rdkafka
+    cyrus_sasl
+  ]
+  ++ lib.optionals withNodeModbus [ libmodbus ]
+  ++ lib.optionals withNodeMqtt [ mosquitto ]
+  ++ lib.optionals withNodeNanomsg [ nanomsg ]
+  ++ lib.optionals withNodeOpenDSS [ opendssc ]
+  ++ lib.optionals withNodeOpalOrchestra [
+    orchestra
+    libxml2
+  ]
+  ++ lib.optionals withNodeRedis [ redis-plus-plus ]
+  ++ lib.optionals withNodeRtp [ libre ]
+  ++ lib.optionals withNodeSocket [ libnl ]
+  ++ lib.optionals withNodeTemper [ libusb1 ]
+  ++ lib.optionals withNodeUldaq [ libuldaq ]
+  ++ lib.optionals withNodeWebrtc [ libdatachannel ]
+  ++ lib.optionals withNodeZeromq [
+    czmq
+    libsodium
+  ];
 
-  propagatedBuildInputs =
-    [
-      libuuid
-      jansson
-    ]
-    ++ lib.optionals withFormatProtobuf [
-      protobuf
-      protobufc
-    ]
-    ++ lib.optionals withNodeInfiniband [ rdma-core ]
-    ++ lib.optionals withExtraConfig [ libconfig ];
+  propagatedBuildInputs = [
+    libuuid
+    jansson
+  ]
+  ++ lib.optionals withFormatProtobuf [
+    protobuf
+    protobufc
+  ]
+  ++ lib.optionals withNodeInfiniband [ rdma-core ]
+  ++ lib.optionals withExtraConfig [ libconfig ];
 
   meta = {
     mainProgram = "villas";


lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+  int _write(struct Sample *smps[], unsigned cnt) override {
+    if (dataDefinitionFileWriteOnly) {
+      logger->warn("Stopping node after writing the DDF file");
+      setState(State::STOPPING);
+      return -1;
+    }
+
+    assert(cnt == 1);
+
+    try {
```
n-eiling, line 306: avoid exceptions

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+                   subscribeMappings);
+    }
+
+    return 0;
+  }
+
+  void checkDuplicateNames(const DataSet &a, const DataSet &b) {
+    for (auto &[name, item] : a.items) {
+      if (b.items.find(name) != b.items.end()) {
+        throw RuntimeError(
```
n-eiling, line 466: avoid exceptions

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+    if (dataDefinitionFilename) {
+      if (!std::filesystem::exists(*dataDefinitionFilename) &&
+          !dataDefinitionFileOverwrite) {
+        throw RuntimeError("OPAL-RT Orchestra Data Definition file (DDF) at "
+                           "'{}' does not exist",
+                           *dataDefinitionFilename);
+      }
+    } else {
+      if (dataDefinitionFileOverwrite) {
+        throw RuntimeError(
```
n-eiling, line 483: avoid exceptions

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+
+    auto ret = dataDefinitionFilename
+                   ? RTConnectWithFile(dataDefinitionFilename->c_str(),
+                                       domain.name.c_str(), connectTimeout)
+                   : RTConnect(domain.name.c_str(), connectTimeout);
+    if (ret != RTAPI_SUCCESS) {
+      throw RTError(ret, "Failed to connect to Orchestra framework");
+    }
+
+    logger->info("Connected to Orchestra framework: domain={}", domain.name);
```
n-eiling, line 527: does this really need to be info?

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+                  connectionKey);
+
+    RTConnectionLockGuard guard(connectionKey);
+
+    auto ret = dataDefinitionFilename
+                   ? RTConnectWithFile(dataDefinitionFilename->c_str(),
+                                       domain.name.c_str(), connectTimeout)
+                   : RTConnect(domain.name.c_str(), connectTimeout);
+    if (ret != RTAPI_SUCCESS) {
+      throw RTError(ret, "Failed to connect to Orchestra framework");
```
n-eiling, line 524: avoid exceptions.

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+  int start() override {
+    if (dataDefinitionFileWriteOnly) {
+      return Node::start();
+    }
+
+    RTConnectionLockGuard guard(connectionKey);
+
+    RTWaitReadyToGo();
+
+    if (!domain.synchronous) {
```
n-eiling, line 589: maybe rename to asynchronous, because I only saw negative checks, and positive checks are more readable.

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+
+    return Node::start();
+  }
+
+  int stop() override {
+    int ret = Node::stop();
+    if (ret)
+      return ret;
+
+    RTConnectionLockGuard guard(connectionKey);
```
n-eiling, line 601: Is this guaranteed to not be reodered? Maybe it's good pratice to encapsulate these guards into scopes (`{...}`)?

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+
+    if (!domain.synchronous) {
+      task.stop();
+    }
+
+    auto ret2 = RTDisconnect();
+    if (ret != RTAPI_SUCCESS) {
+      throw RTError(ret2, "Failed to disconnect");
```
n-eiling, line 609: this function is even designed to use int to signal sucess, so here you really shouldnt use exceptions

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+    return (int)NodeFactory::Flags::SUPPORTS_READ |
+           (int)NodeFactory::Flags::SUPPORTS_WRITE |
+           (int)NodeFactory::Flags::SUPPORTS_POLL |
+           (int)NodeFactory::Flags::HIDDEN;
+  }
+
+  std::string getName() const override { return "opal.orchestra"; }
+
+  std::string getDescription() const override {
+    return "OPAL-RT Orchestra client";
```
n-eiling, line 675: could be a bit more informative.

lib/nodes/orchestra/ddf.cpp
```diff
@@ -0,0 +1,410 @@
...
+  xmlNewProp(node, BAD_CAST "name", BAD_CAST name.c_str());
+  xmlNewChild(node, NULL, BAD_CAST "type", BAD_CAST "bus");
+
+  for (const auto &p : items) {
+    auto &item = p.second;
+    item->toXml(node, withDefault);
+  }
+}
+
+DataItem *BusItem::upsertItem(std::vector<std::string> pathComponents,
```
n-eiling, line 74: could use a more descriptive name. What is upsert?

lib/nodes/orchestra/ddf.cpp
```diff
@@ -0,0 +1,410 @@
...
+      "extcomm", &ec, "addrframework", &af, "portframework", &pf,
+      "nicframework", &nf, "nicclient", &nc, "coreframework", &cf, "coreclient",
+      &cc);
+  if (ret) {
+    throw ConfigError(json, err, "node-config-node-opal-orchestra-connection",
+                      "Failed to parse configuration of local connection");
+  }
+
+  if (ec) {
+    extcomm = ec;
```
n-eiling, line 255: why not use variable directly in unpack?

lib/nodes/orchestra/ddf.cpp
```diff
@@ -0,0 +1,410 @@
...
+              BAD_CAST fmt::format("{}", pciIndex).c_str());
+}
+
+void ConnectionRemote::parse(json_t *json) {
+  const char *nme;
+
+  json_error_t err;
+
+  int ret = json_unpack_ex(json, &err, 0, "{ s: s, s: i }", "card", &nme,
+                           "pciindex", &pciIndex);
```
n-eiling, line 298: use proper snake case (`pci_index`)

lib/nodes/orchestra/ddf.cpp
```diff
@@ -0,0 +1,410 @@
...
+  xmlNewProp(connection, BAD_CAST "nodeIdFramework",
+             BAD_CAST fmt::format("{}", nodeIdFramework).c_str());
+  xmlNewProp(connection, BAD_CAST "segmentId",
+             BAD_CAST fmt::format("{}", segmentId).c_str());
+}
+
+void ConnectionDolphin::parse(json_t *json) {
+  json_error_t err;
+  auto ret = json_unpack_ex(json, &err, 0, "{ s: i, s: i }", "nodeIdFramework",
+                            &nodeIdFramework, "segmentId", &segmentId);
```
n-eiling, line 328: use snake case

lib/nodes/orchestra/ddf.cpp
```diff
@@ -0,0 +1,410 @@
...
+  if (strcmp(type, "local") == 0) {
+    c = new ConnectionLocal();
+    c->parse(json);
+  } else if (strcmp(type, "remote") == 0) {
+    c = new ConnectionRemote();
+    c->parse(json);
+  } else if (strcmp(type, "dolphin") == 0) {
+    c = new ConnectionDolphin();
+    c->parse(json);
+  } else {
```
n-eiling, line 356: consider using a smart pointer and removing the exceptions

lib/nodes/orchestra/locks.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include <mutex>
+
+extern "C" {
+#include <RTAPI.h>
+}
+
+#include <villas/nodes/orchestra/locks.hpp>
+
+static std::mutex globalLock;
```
n-eiling, line 17: this looks like a connection lock, but is implemented as a global lock. Is this intentional? Can we at least add a note that the lock should in the future be moved to a connection object?

lib/nodes/orchestra/signal.cpp
```diff
@@ -0,0 +1,221 @@
...
+  } else if (t == "int64") {
+    return orchestra::SignalType::INT64;
+  } else if (t == "float32") {
+    return orchestra::SignalType::FLOAT32;
+  } else if (t == "float64") {
+    return orchestra::SignalType::FLOAT64;
+  } else if (t == "bus") {
+    return orchestra::SignalType::BUS;
+  } else {
+    throw RuntimeError("Unknown Orchestra signal type: {}", t);
```
n-eiling, line 82: You could return an optional instead, or result if C++ has that.

lib/nodes/orchestra/signal.cpp
```diff
@@ -0,0 +1,221 @@
...
+  case orchestra::SignalType::INT64:
+    return "int64";
+  case orchestra::SignalType::FLOAT32:
+    return "float32";
+  case orchestra::SignalType::FLOAT64:
+    return "float64";
+  case orchestra::SignalType::BUS:
+    return "bus";
+  default:
+    throw RuntimeError("Unknown Orchestra signal type: {}", (int)t);
```
n-eiling, line 114: avoid exception

lib/nodes/orchestra/signal.cpp
```diff
@@ -0,0 +1,221 @@
...
+  }
+}
+
+SignalData
+villas::node::orchestra::toNodeSignalData(const char *orchestraData,
```
n-eiling, line 119: if `orchestraData` is opaque data, shouldn't it be `const void*`? Also a good candidate for a smart pointer.

n-eiling: This function is also not checking the size or if `orchestraData` is nullptr

CMakeLists.txt
```diff
@@ -158,55 +160,56 @@ else()
...
+cmake_dependent_option(WITH_NODE_INFLUXDB       "Build with influxdb node-type"                         "${WITH_DEFAULTS}" "" OFF)
+cmake_dependent_option(WITH_NODE_KAFKA          "Build with kafka node-type"                            "${WITH_DEFAULTS}" "RDKAFKA_FOUND" OFF)
+cmake_dependent_option(WITH_NODE_LOOPBACK       "Build with loopback node-type"                         "${WITH_DEFAULTS}" "" OFF)
+cmake_dependent_option(WITH_NODE_MODBUS         "Build with modbus node-type"                           "${WITH_DEFAULTS}" "MODBUS_FOUND" OFF)
+cmake_dependent_option(WITH_NODE_MQTT           "Build with mqtt node-type"                             "${WITH_DEFAULTS}" "MOSQUITTO_FOUND" OFF)
+cmake_dependent_option(WITH_NODE_NANOMSG        "Build with nanomsg node-type"                          "${WITH_DEFAULTS}" "NANOMSG_FOUND" OFF)
+cmake_dependent_option(WITH_NODE_NGSI           "Build with ngsi node-type"                             "${WITH_DEFAULTS}" "" OFF)
+cmake_dependent_option(WITH_NODE_OPAL_ORCHESTRA "Build with the opal-orchestra node-type"               "${WITH_DEFAULTS}" "OpalOrchestra_FOUND; LibXml2_FOUND" OFF)
```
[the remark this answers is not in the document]
stv0g: The FOUND variables are set by [`find_package_handle_standard_args`](https://cmake.org/cmake/help/latest/module/FindPackageHandleStandardArgs.html), which prefixes them by a package name. By convention package names in CMake are written in Pascal case.

See here for examples in the CMake documentation: https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html

etc/examples/nodes/opal_orchestra.conf
```diff
@@ -0,0 +1,106 @@
...
+        # Enable to overwrite the DDF file.
+        # This is useful when you want to generate the DDF file from the configuration file
+        # for importing it into RT-LAB or HYPERSIM.
+        ddf_overwrite = true
+
+        # Orchestra domain name.
+        domain = "domain1"
+
+        # Connection timeout in seconds.
+        connect_timeout = 2
```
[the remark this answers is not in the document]
stv0g: I am not a big fan if encoding units, data-types or alike in the setting name.
But I agree that using different scales (microseconds vs seconds is not optimal).

I would like to extend the configuration parsing globally to also accept strings to these settings which allows parsing durations with units.
E.g. "4s", "10us".

etc/examples/nodes/opal_orchestra.conf
```diff
@@ -0,0 +1,106 @@
...
+
+        # Define the delay to wait when using flag synchronisation (XHP).
+        # This will call the system function usleep and free the CPU.
+        # In micro-seconds.
+        flag_delay = 0
```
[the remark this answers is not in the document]
stv0g: Same as above.

etc/examples/nodes/opal_orchestra.conf
```diff
@@ -0,0 +1,106 @@
...
+        states = false
+        multiple_publish_allowed = false
+
+        connection = {
+            # One of: local, remote or dolphin
+            type = "local"
+
+            # For 'local'
+            extcomm = "udp";
+            addrframework = "10.168.13.5";
```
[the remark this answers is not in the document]
stv0g: > Maybe not a good idea to include some internal IP here?

Good point. I will change this.

> Also consider using proper snake case (addr_framework), same for all keys below.

These settings originate from our Orchestra software and are passed verbatim.
Hence, I would prefer to keep the same naming to avoid confusion.

See: https://opal-rt.atlassian.net/wiki/spaces/PRD/pages/143626248/Orchestra

etc/examples/nodes/opal_orchestra.conf
```diff
@@ -0,0 +1,106 @@
...
+            coreclient = 0
+            nicframework = "eno2"
+            nicclient = "eno1"
+
+            # For 'remote'
+            card = "test"
+            pciindex = 0
+
+            # For 'dolphin'
+            nodeIdFramework = 0
```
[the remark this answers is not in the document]
stv0g: As above.

include/villas/nodes/orchestra/error.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * SPDX-FileCopyrightText: 2025, OPAL-RT Germany GmbH
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <string>
+
+extern "C" {
+#include <RTAPI.h>
```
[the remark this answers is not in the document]
stv0g: Yeah, indentation here is removed by clang-format.

lib/nodes/orchestra/ddf.cpp
stv0g: See above. These are external settings.

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+        logger->warn("Orchestra Framework has been stopped / unmapped");
+        setState(State::STOPPING);
+      } else {
+        logger->error("Failed to read from Orchestra: {}", e.what());
+      }
+      return -1;
+    }
+  }
+
+  int _write(struct Sample *smps[], unsigned cnt) override {
```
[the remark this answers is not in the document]
stv0g: This is not a change introduced by this PR. Its the interface of `villas::node::Node`

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+        &sw, "ddf_overwrite", &ow, "ddf_overwrite_only", &owo, "rate", &rate,
+        "in", "signals", &json_in_signals, "out", "signals", &json_out_signals);
+    if (ret) {
+      throw ConfigError(json, err, "node-config-node-opal-orchestra");
+    }
+
+    domain.name = dn;
+
+    if (ddf) {
+      dataDefinitionFilename = ddf;
```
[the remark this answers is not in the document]
stv0g: `dataDefinitionFilename` is `std::string` which the C-library jansson can not directly unpack.

lib/nodes/orchestra.cpp
stv0g: This is an external setting. I would like to keep the semantics aligned with the Orchestra software here.

lib/nodes/orchestra/signal.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+  case orchestra::SignalType::FLOAT64:
+    return *reinterpret_cast<const double *>(orchestraData);
+
+  default:
+    throw RuntimeError("Orchestra signal type {} is not supported",
+                       signalTypeToString(orchestraType));
+  };
+
+  return SignalData(); // Unreachable
```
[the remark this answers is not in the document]
stv0g: I believe we need a return here. C++23 has [`std::unreachable`](https://en.cppreference.com/w/cpp/utility/unreachable.html). But we are still on C++17, so we can not use that..

etc/examples/nodes/opal_orchestra.conf
```diff
@@ -0,0 +1,106 @@
+# SPDX-FileCopyrightText: 2025, OPAL-RT Germany GmbH
+# SPDX-License-Identifier: Apache-2.0
+
+stats = 1
+
+nodes = {
+    domain1 = {
+        type = "opal.orchestra"
+
+        # Path  to the OPAL-RT Orchestra Data Defintion XML file (DDF).
```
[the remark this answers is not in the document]
stv0g: Fixed

lib/nodes/orchestra/signal.cpp
stv0g: > if orchestraData is opaque data, shouldn't it be const void*?

Fixed

lib/nodes/orchestra/signal.cpp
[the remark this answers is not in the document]
stv0g: I added a better comment.

lib/nodes/orchestra/signal.cpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+void villas::node::orchestra::toOrchestraSignalData(
+    char *orchestraData, orchestra::SignalType orchestraType,
```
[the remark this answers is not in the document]
stv0g: Fixed

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+
+    auto logger = Log::get("orchestra");
+    logger->trace(
+        "Prepared mapping: path='{}', type={}, typeSize={}, key={}, buffer={}"
+        "length={}, default={}",
+        path, orchestra::signalTypeToString(item->type), key, buffer, typeSize,
+        length, item->defaultValue);
+  }
+
+  void publish(struct Sample *smp) {
```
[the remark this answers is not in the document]
stv0g: Fixed

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+      }
+
+      toOrchestraSignalData(orchestraDataPtr, item->type, smp->data[index],
+                            signal->type);
+
+      orchestraDataPtr += typeSize;
+    }
+  }
+
+  void subscribe(struct Sample *smp) {
```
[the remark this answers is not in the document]
stv0g: Fixed

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+  unsigned int
+      flagDelay; // Define the delay to wait, this will call the system function usleep and free the CPU. In microseconds.
+  bool
+      useFlagWithTool; // Force the local Orchestra communication to be made with flag instead of semaphore when using an external communication process.
+  bool skipWaitToGo;   // Skip wait-to-go step during start.
+  bool dataDefinitionFileOverwrite; // Overwrite the data definition file (DDF).
+  bool
+      dataDefinitionFileWriteOnly; // Overwrite the data definition file (DDF) and terminate VILLASnode.
+
+  int _read(struct Sample *smps[], unsigned cnt) override {
```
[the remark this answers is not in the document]
stv0g: This is not a change introduced by this PR. Its the interface of villas::node::Node.

I am open to change this in another follow-up PR.

lib/nodes/orchestra/ddf.cpp
stv0g: Update is a commonly used term in data base to describe a transaction which either **up**dates or **in**serts an entry depending on wether it already exists.

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+  std::vector<Signal::Ptr> signals;
+
+  // Cached signal indices
+  // We keep a vector of indices to map the signal index in the signal list.
+  SignalList::Ptr signalList; // Signal list for which the indices are valid.
+  std::vector<int> indices;
+
+  // Run-time members which will be retrieved from Orchestra in prepare().
+  unsigned short key;
+  char *buffer;
```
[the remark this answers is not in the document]
stv0g: This buffer is managed by Orchestra RTAPI. So RAII does not make sense here. See `RTGetBufferInfo()`.

etc/examples/nodes/opal_orchestra.conf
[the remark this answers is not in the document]
n-eiling: sounds good

CMakeLists.txt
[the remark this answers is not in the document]
n-eiling: Every other variable is using a different style, though....

lib/nodes/orchestra.cpp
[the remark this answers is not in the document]
n-eiling: ah yes I remember.

CMakeLists.txt
```diff
@@ -76,6 +76,8 @@ find_package(Lua)
 find_package(LibDataChannel)
 find_package(re)
 find_package(OpenDSSC)
+find_package(OpalOrchestra)
```
al3xa23, line 79: Maybe I missed it, but the libraries are not installed in deps. Does it come later?

etc/examples/nodes/opal_orchestra.conf
```diff
@@ -0,0 +1,106 @@
...
+    domain1 = {
+        type = "opal.orchestra"
+
+        # Path to the OPAL-RT Orchestra Data Defintion XML file (DDF).
+        ddf = "orchestra.xml"
+
+        # Enable to overwrite the DDF file.
+        # This is useful when you want to generate the DDF file from the configuration file
+        # for importing it into RT-LAB or HYPERSIM.
+        ddf_overwrite = true
```
al3xa23, line 16: overwriteonly is defined in orchestra.cpp. Maybe you can add it here to with description when to use (not clear to me yet)

packaging/nix/orchestra.nix
```diff
@@ -0,0 +1,48 @@
...
+  dpkg,
+  libuuid,
+  makeWrapper ,
+}:
+stdenv.mkDerivation {
+  pname = "libOpalOrchestra";
+  version = "7.6.2";
+
+  src = fetchurl {
+    url = "https://blob.opal-rt.com/softwares/rt-lab-archives/componentorchestra_7.6.2_amd64.deb?sp=r&st=2024-10-30T06:31:59Z&se=2034-11-30T14:31:59Z&spr=https&sv=2022-11-02&sr=b&sig=cnKY8RxZf8hv91gWLIBG6iBGSVziXkKR3%2BOYIE6MSkI%3D";
```
al3xa23, line 19: Is it intentional that only nix downloads the library? What about deps.sh?

etc/examples/nodes/opal_orchestra.conf
```diff
@@ -0,0 +1,106 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+stats = 1
+
+nodes = {
+    domain1 = {
+        type = "opal.orchestra"
+
+        # Path to the OPAL-RT Orchestra Data Defintion XML file (DDF).
+        ddf = "orchestra.xml"
```
al3xa23, line 11: Are you planning to provide the xml file in the repo?

al3xa23 (commented):
Hi, nice work. I added some minor comments since I am not yet that familiar with code reviews. 
General question: I assume the tests are following in the next PR?

CMakeLists.txt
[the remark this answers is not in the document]
stv0g: Yes, unfortuantely. However, the package names are defined by the third-party packages themselves, Hence, we have no authority for changing them.

lib/nodes/orchestra.cpp
stv0g: I am not sure if I can explain this. But I would say this is pretty common practice of using RAII for locking (see `std::lock_guard`).

Here is also a StackOverflow article which explains it with references to the C++ standard:

> Those things [the RTConnectionLockGuard] that you're describing (the function call and the instantiation of the lock) are known as full-expressions in the standard.
>
> As per C++11 1.9 Program execution /14 (same location and text in C++14, same text in C++17 4.6 Program execution /16):
>
> >    Every value computation and side effect associated with a full-expression is sequenced before every value computation and side effect associated with the next full-expression to be evaluated.
>
> There are cases where seemingly sequential operations can be indeterminately sequenced, but this is not one of them.

(https://stackoverflow.com/a/48512542)

lib/nodes/orchestra.cpp
```diff
@@ -0,0 +1,679 @@
...
+
+  OrchestraMapping(DataItem *item, const std::string &path)
+      : item(item), path(path), signals(), signalList(), indices() {}
+
+  void addSignal(Signal::Ptr signal, int orchestraIdx) {
+    if (orchestraIdx < 0) {
+      orchestraIdx = signals.size();
+    }
+
+    if (orchestraIdx < static_cast<int>(signals.size())) {
```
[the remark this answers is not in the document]
stv0g: Because I would otherwise compare integers of different signedness:

```
/home/stv0g/workspace/villas/node/lib/nodes/orchestra.cpp:66:22: error: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<std::shared_ptr<villas::node::Signal> >::size_type’ {aka ‘long unsigned int’} [-Werror=sign-compare]
   66 |     if (orchestraIdx < signals.size()) {
      |         ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
```

Ideally, I would convert orchestraIdx to be unsigned as well.
However, I committed a small crime and reused the special value `-1` to indicate the absence of an index.
I should rewrite this with `std::optional` instead.

n-eiling:
Would be interesting to know what the license states specifically. Can we even publish this as open source? If yes, then at least the header needs to be open source as well, right? Or is there some clause specifying terms for use as a library?

stv0g:
> Would be interesting to know what the license states specifically. Can we even publish this as open source? If yes, then at least the header needs to be open source as well, right? Or is there some clause specifying terms for use as a library?

The libOpalOrchestra library and its header which are not licensed under an open-source license. This contribution includes neither the library itself or its header.

It merely uses it by dynamically linking against it.

I do not see an issue here as we are licensing the node-type itself (this contribution) under a permissive OSS license (Apache-2.0).

PR #926, opened 2025-06-17
Change method of python api restart node request to POST instead of default GET

SystemsPurge:
Otherwise results in
```
07:27:46 warn api:session API request failed: endpoint=restart, method=GET, code=400: The 'restart' API endpoint does not support GET requests
```
In villasnode
Maybe it is recommended to just use load_config instead and remove this entirely ?


stv0g:
LGTM. Thanks :)

PR #933, opened 2025-06-26
Fix #678: Fix stats print on ARM due to indeterminate state of "va_list" parameter after vasprintf

### LINKED ISSUE #678: Broken stats on aarch64 release build
An aarch64 release build may cause the stats to show `std::numeric_limits<long int>::max` for any data. Example log output:
```
11:13:10 info stats             Node           │           Recv │           Sent │           Drop │           Skip │       OWD last │       OWD mean │      Rate last │      Rate mean │       Age mean │        Age Max │     Signals
11:13:10 info stats                            │           pkts │           pkts │           pkts │           pkts │           secs │           secs │        pkt/sec │        pkt/sec │           secs │            sec │     signals
11:13:10 info stats            ────────────────┼────────────────┼────────────────┼────────────────┼────────────────┼────────────────┼────────────────┼────────────────┼────────────────┼────────────────┼────────────────┼─────────────
11:13:11 info stats             webrtc         │ 28147402498383 │ 28147402498383 │ 28147402498383 │ 28147402498383 │       0.013749 │       0.013749 │       0.013749 │       0.013749 │       0.013749 │       0.013749 │ 28147402498 
11:13:12 info stats             webrtc         │ 28147402498383 │ 28147402498383 │ 28147402498383 │ 28147402498383 │       0.014313 │       0.014313 │       0.014313 │       0.014313 │       0.014313 │       0.014313 │ 28147402498 
11:13:13 info stats             webrtc         │ 28147402498383 │ 28147402498383 │ 28147402498383 │ 28147402498383 │       0.014452 │       0.014452 │       0.014452 │       0.014452 │       0.014452 │       0.014452 │ 28147402498 
11:13:14 info stats             webrtc         │ 28147402498383 │ 28147402498383 │ 28147402498383 │ 28147402498383 │       0.013672 │       0.013672 │       0.013672 │       0.013672 │       0.013672 │       0.013672 │ 28147402498 
11:13:15 info stats             webrtc         │ 28147402498383 │ 28147402498383 │ 28147402498383 │ 28147402498383 │       0.013599 │       0.013599 │       0.013599 │       0.013599 │       0.013599 │       0.013599 │ 28147402498 
11:13:16 info stats             webrtc         │ 28147402498383 │ 28147402498383 │ 28147402498383 │ 28147402498383 │       0.014081 │       0.014081 │       0.014081 │       0.014081 │       0.014081 │       0.014081 │ 28147402498 
```

IgnoreWarnings:
The state of "va_list" parameter is indeterminant after a call to vasprintf, which causes UB on ARM.
For detailed error explanation: https://github.com/VILLASframework/node/issues/678.

1. Portability: Specifiying the "va" parameter as reference to ensure reproducable behavior on arm and x86.
2. "vstrcatf" now copies the va_list parameter to prevent it beeing undetermined after call to vasprintf. 

@stv0g 
I made the decision to put the responsibility of advancing the list in the called function rather than the caller because this is how the existing code in table.cpp:void Table::row(int count, ...) assumes it should work.

@fwege

### DIFF
--- a/common/lib/utils.cpp
+++ b/common/lib/utils.cpp
@@ -197,7 +197,21 @@ double randf() { return (double)random() / RAND_MAX; }
 char *vstrcatf(char **dest, const char *fmt, va_list &ap) {
   char *tmp;
   int n = *dest ? strlen(*dest) : 0;
-  int i = vasprintf(&tmp, fmt, ap);
+
+  va_list va_copy;
+  va_copy(va_copy, ap);
+  int i = vasprintf(&tmp, fmt, va_copy);
+  va_end(va_copy);
+
+  // Advance va_list
+  const char *format = fmt;
+  if (strcmp(format, "%ju") == 0) {
+    va_arg(ap, unsigned int);
+  } else if (strcmp(format, "%lf") == 0) {
+    va_arg(ap, double);
+  } else if (strcmp(format, "%s") == 0) {
+    va_arg(ap, char *);
+  }
 
   *dest = (char *)(realloc(*dest, n + i + 1));
   if (*dest != nullptr)


common/lib/utils.cpp
```diff
@@ -194,10 +194,24 @@ double boxMuller(float m, float s) {
 
 double randf() { return (double)random() / RAND_MAX; }
 
-char *vstrcatf(char **dest, const char *fmt, va_list ap) {
+char *vstrcatf(char **dest, const char *fmt, va_list &ap) {
   char *tmp;
```
n-eiling, line 198: I would add null pointer checks for `dest` and `fmt`

common/lib/utils.cpp
```diff
@@ -194,10 +194,24 @@ double boxMuller(float m, float s) {
 
 double randf() { return (double)random() / RAND_MAX; }
 
-char *vstrcatf(char **dest, const char *fmt, va_list ap) {
+char *vstrcatf(char **dest, const char *fmt, va_list &ap) {
```
n-eiling, line 197: I'm surprised you don't need `const char* restrict fmt`.

n-eiling (approved):
LGTM, some minor comments above.

common/lib/utils.cpp
stv0g: To my knowledge `restrict` is a C not C++ keyword?

Are you concerned about aliasing between the `*dest` and `fmt` pointers?

I dont fully understand, why we might need it..

common/lib/utils.cpp
```diff
@@ -194,10 +194,24 @@ double boxMuller(float m, float s) {
...
+
+  // Advance va_list
+  const char *format = fmt;
+  if (strcmp(format, "%ju") == 0) {
+    va_arg(ap, unsigned int);
+  } else if (strcmp(format, "%lf") == 0) {
+    va_arg(ap, double);
+  } else if (strcmp(format, "%s") == 0) {
+    va_arg(ap, char *);
+  }
```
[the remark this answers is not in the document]
stv0g: I am wondering, isnt it also UB to call `vsnprintf` again with the same `va_copy`?

@IgnoreWarnings wrote in #678

> This bug occures because "va_list" is used after beeing passed to "vasprintf", which leads to undeterministic behavior.
Functions like "vasprintf" "invoke va_arg at least once, the value of arg is indeterminate after the return." (https://en.cppreference.com/w/c/io/vfprintf : Notes).

stv0g:
Hi @IgnoreWarnings,

I think a found a nicer solution to the issue in #935.

What do you think about it? Thanks for the analysis. I definitely learned something new :)

PR #934, opened 2025-06-26
Add: miob.conf example for platform based fpgas, Restored #879

IgnoreWarnings:
Add example config for miob.
Restored: #879


stv0g:
@IgnoreWarnings Is there anything left, which blocks the PR? The CI is passing already..

IgnoreWarnings:
> Hi @IgnoreWarnings, @n-eiling,
> 
> I am wondering, do you still forsee progress on this? I am uncertain how Niklas's departure from ACS might have affected this work?
> 
> Should we keep this PR, or close it?

TLDR:
Fix that the miob config test gets ignored and merge.

Hi, this PRs content is complete, but fails the pipeline because ignoring the test doesnt work properly. The config works on hw but due to bad villas code makes the parsing require a vfio hw dependency that doesnt work in ci. I discussed this in more detail in #879 (restored). As you can see in the commits, ignoring the test didnt work somehow. Maybe you can fix this.

IgnoreWarnings:
I may just have found the error, lets see if ci happy.

IgnoreWarnings:
@stv0g do you know why the ci pre commit fails with end of file?

PR #939, opened 2025-07-03
Update license note

stv0g:
This PR updates the license note in the projects README file as requested in this review:

https://github.com/VILLASframework/node/issues/936#issuecomment-3028316220

PR #944, opened 2025-07-09
Use "The VILLASframework Authors" as copyright owners in LICENSE

n-eiling:
Let's continue the discussion from #939 here.

Pro:
  - Have a copyright owner in the LICENSE file
  - It is broad enough to include everybody
  - This is recommended by the [Linux Foundation](https://www.linuxfoundation.org/blog/blog/copyright-notices-in-open-source-software-projects)
  - Was requested by one of the JOSS Reviewers
 
Con:
  - Generic LICENSE files are acceptable according to SPDX (I think?), so this may be a change without a need
  - We have to update the copyright year every year.
  - Inconsistent with copyright notices in files (although it is not contradicting those)

In the future we could also use "The VILLASframework Authors" in files missing more explicit copyright notices.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ project(villas-node
 
 # Some more project settings
 set(PROJECT_AUTHOR "Steffen Vogel")
-set(PROJECT_COPYRIGHT "2014-2021, Institute for Automation of Complex Power Systems, RWTH Aachen University")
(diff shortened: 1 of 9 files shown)


stv0g:
I think this is a good change 👍🏻

Should we also update the copyright / author notices in the licenses located in the `LICENSES` folder?

n-eiling:
Added it there and also found some more locations that seemed sensible.

stv0g (approved):

PR #952, opened 2025-08-21
fix(compat): Add drop-in std::filesystem header for older systems

stv0g:
Needed for some older compilers which do not include std::filesystem by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,19 +76,16 @@ find_package(Lua)
 find_package(LibDataChannel)
 find_package(re)
 find_package(OpenDSSC)
+find_package(FileSystem)
 
 # Check for tools
+find_program(PROTOBUFC_COMPILER NAMES protoc-c)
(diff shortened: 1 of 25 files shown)


stv0g:
> Looks in the CI as if ghc_filesystem is the default? Shouldn't we prefer std::filesystem?

Dang, I've implement a CMake check to see if the library is available.. However, that does not seem to work.. I will check.

> Also, do you think it's a good idea to include the third party code as a copy in this repo?

I agree. But I am not sure how we want to implement this? Git submodules? CMake's [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html)? Or via out `deps.sh` shell script?

stv0g:
I've removed the external header file. Instead we now rely on CMake's find_package() for finding the header-only library (ghc::filesystem has good CMake support for it to work flawlessly).

n-eiling (dismissed):
Looks great :)

PR #955, opened 2025-09-02
Fix RTP node compilation with libre 3.24

pjungkamp:
This bug was introduced by https://github.com/baresip/re/pull/1357

### DIFF
--- a/include/villas/nodes/rtp.hpp
+++ b/include/villas/nodes/rtp.hpp
@@ -24,6 +24,7 @@ extern "C" {
 // clang-format off
 #include <re/re_types.h>
 #include <re/re_sa.h>
+#include <re/re_list.h>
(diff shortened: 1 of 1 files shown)


stv0g (dismissed):
Just a tiny nit

stv0g:
This fix is also included in https://github.com/VILLASframework/node/pull/954.
But I will merge your PR first and rebase #954 after.

pjungkamp:
I just removed the comment entirely.

`libre` doesn't use include guards and doesn't declare transitive dependencies in any of their headers. I don't like it but this seems to be the way it's intended to be used and thus not-a-bug in `libre`.

stv0g:
As a library user we should probably only include `re/re.h` which has the guard.

We are not doing it due a colliding identifier with libjansson.

PR #956, opened 2025-09-02
Enable C++20 support and fix all formatting errors detected by `consteval`

pjungkamp:
This is based on and includes #955.


stv0g:
I'd also like to use the new features of C++20. So kets find a way we can make this work with the OPAL-RT targets.

The OPAL-RTLinux OS ships with a GCC v8.3 if I am not mistaken.

Hence we either need to install another more recent toolchain or use Nix to build everything.

Ideally, I would like to build everything with Nix. However there some local dependencies (headers and static libraries on which we rely.

Do you know a way to package those local files into a Nix derivation?

There is `requestFile`...

n-eiling:
Shouldn't we keep a CI target that tests C++17 compatibility, or do you want to drop this? If yes, we would need to ask all users to test this on their setups, no? I don't think the advantages merit breaking some people's setups.

stv0g:
I agree with @pjungkamp on this. For the OPAL-RT targets, I figured out a way to work around the older compiler by using nix. Installing a separate tool chain should probably work there as well.

stv0g:
@pjungkamp Could you approve #954? After this we rebase this one to fix the RockyLinux build error.

pjungkamp:
> Well, we are actually targeting some embedded device with a custom toolchain stuck on an old compiler: [MIOB](https://github.com/RWTH-ACS/miob/tree/main/sw/yocto) It's based on Yocto Gatesgarth that uses clang 11.1/gcc 10.2. 

Yocto 3.2 (Gatesgarth) has been [EOL for 4 years now](https://wiki.yoctoproject.org/wiki/Releases), it was already EOL when you pushed the [initial Yocto layer in 2022](https://github.com/RWTH-ACS/miob/commit/b150e65ddc0f1083a4d22b30075eaadad19a4181). The oldest non-EOL Yocto is 4.0 LTS (kirkstone) which ships with [GCC 11](https://git.yoctoproject.org/poky/tree/meta/recipes-devtools/gcc/gcc-11.5.inc?h=kirkstone&id=c6cd61f1a22d04ba3241e00caf6f35cc2ca76da1). Is there anyone maintaining MIOB? Are you updating VILLASnode and nothing else?

---

> Also we support OPAL-RT with it's weird toolchain and raspbian (We probably want to keep support for quite old RPis (v2?).

The OPAL-RT targets were addressed by @stv0g and the current (non-legacy) Raspberry Pi OS supporting all Raspberry Pis is based on Debian Bookworm and thus includes GCC 11.

n-eiling:
Well, it worked in February. There haven't been major changes to the code related to FPGAs and the CI runs using C++17, so I don't see any reason it should not still work.

I think just because we did a bad job keeping compatibility in the past should not be a reason to now break stuff, or do you disagree?
Regarding effort: This PR puts in effort to stop support for C++17 - not merging this PR would not break anything.
When we wanted to use C++20 features in the past, we always put in alternatives for C++17 and this PR stops with this.

I think for any breaking change we should make sure
1. The change and a migration path are clearly documented.
2. We don't break any use case of people actually using VILLASnode.
3. We don't introduce any regressions.

I think this PR fulfills 3. As I said, I'm not sure about 2. (depends on the state of MIOB and the use of Raspbian Legacy). I don't think this PR fulfills 1.

Just my two cents. Feel free to ignore if you don't share my desire for more long-term stability.

n-eiling:
I mean, I'm not using MIOB anymore and I have no way of testing on a MIOB. Thats why I tagged @fuentesgrau. There are instructions on the yocto website on how to build a yocto distribution. If anyone at ACS could just rebuild and upload the latest SDK it should also be easy to just use that for a CI. I'd be willing to set the CI up.

stv0g (approved):

PR #959, opened 2025-09-05
pre-commit: Add DCO check

pjungkamp:
I found myself force-pushing some commits in **every** PR I've made, just to reword and signoff on all commits. I wrote a little prepare-commit-msg hook that appends the DCO to every commit message automatically.

---

You can install this hook using:

```shell
pre-commit install -t prepare-commit-msg
```

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -2,3 +2,5 @@
 # SPDX-License-Identifier: Apache-2.0
 
 use flake
+pre-commit install
+pre-commit install -t prepare-commit-msg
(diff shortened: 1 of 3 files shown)


stv0g (approved):

n-eiling:
This is not a check, but it adds the sign-off automatically, right? Why don't you use [`format.signOff`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-formatsignOff)?

stv0g:
I believe there are many pre-commit checks which are "auto-fixing" the code.

This would be one of them. Contributors could still use `format.signOff`, and this "check" would boild down to no-op.

PR #961, opened 2025-09-08
Add RockyLinux 10 builds to CI

stv0g:


n-eiling:
Why are you doing this? I think Rocky 9 might still be used at ACS, or not? I wouldn't remove a distribution that is actively being used for running VILLASnode.

PR #965, opened 2025-10-26
Add monitoring hook

Oussema-ben-h:
This Hook replaces the values of the signals by the metrics: Rate, Delay, lost and out-of-order sequences

### DIFF
--- a/lib/hooks/CMakeLists.txt
+++ b/lib/hooks/CMakeLists.txt
@@ -6,6 +6,7 @@
 
 set(HOOK_SRC
     average.cpp
+    monitoring_hook.cpp
     cast.cpp
     decimate.cpp
     digest.cpp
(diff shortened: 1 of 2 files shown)


stv0g:
Dear @Oussema-ben-h,

I beliefe this functionality is already present in the current version of VILLASnode.

We support using special signal expressions in the path configuration:
https://villas.fein-aachen.org/docs/node/config/paths#complex-signal-mapping-expressions

### Example

```
paths = [
	{
		in = [
                        "udp_node.data[0-2]",
			"udp_node.hdr.sequence",
			"udp_node.hdr.length",
			"udp_node.ts.origin",
			"udp_node.ts.received",
			"udp_node.stats.owd.last",
			"udp_node.stats.skipped.total",
		],
		...
	}
]
```

al3xa23:
We wanted to add some extra functionalities to make the montioring easier. This hook also calculates lost and out-of-order packets and not only skipped packets. Do you think it makes sense to include it here?

stv0g:
Hi @al3xa23,

The approach, which I mentioned above, already exposes all available statistics (including out-of-order, and skipped packets).

al3xa23:
Okay, then it seems we don't need it.

PR #990, opened 2026-01-15
Add new gateway node-type

JitpanuM:
Add new gateway node-type for transforming an api type comming to VILLASnode http server.
Currently it only support transformation of http and gRPC.
The protobuf payload that is not VILLASnode is also support using gRPC server reflection.
The gRPC server reflection feature is not available when install gRPC with package manager.
The additonal script is made to get the reflection.proto file from gRPC repo.


al3xa23:
Hi @stv0g, we chose originally gateway because we want to support different API conversions/translations. However, if you think, it makes more sense to have it in one API node-type, we can change it of couse.

stv0g:
Hi @al3xa23 @JitpanuM,

I think it makes sense to just name it "api". As far as I understand, we are not talking about a standardized API here. Its rather a custom VILLASnode API available over some network transports (HTTP, gRPC, ...).

For the existing node-type, its also the case. It was a failed attempt at creating a unfied API in the ERIGrid project.

So I would like to get rid of this node-type to reduce the maintenance burden.

As long as all these transport share the same foundation (methods, datatypes, ...) we should aggregate them in the same node-type.

However, we also support other APIs like NGSI via the `ngsi` node-type.
In this case its a standardized API, which should deserve its own node-type as its using different methods, data-types etc..

Going forward, I propose:

- Step 1: We create and merge a PR to remove the old `api` node-type
- Step 2: You rename the node-type in this PR to `api`
- Step 3: We rebase and merge this PR

What do you think? I am interested in your ideas too :)

al3xa23:
Hi, sounds reasonable to me. We can do it like this :)

riteshk001:
After rebasing branch node-gateway with master (this removes all of the old node-type API implementation), steps to rename node-gateway to node-api: 

- [x] Rename source files gateway.cpp/hpp to api.cpp/hpp.
- [x] Refactor code in gateway.cpp/hpp to use ApiGatewayNode instead of GatewayNode for clarity.
- [x] Migrate OpenAPI yaml files from gateway.yaml to api.yaml.
- [x] Rename example conf file to use node name `api` instead of `gateway`.
- [ ] Change HTTP handler `/api/v2/gateway/...` (clarity required whether to keep existing path or to change it). 

I am not sure whether the HTTP handler should be kept the same or it should be changed to something like `api/v2/api/...` although that includes a redundant `api`. 

If there are any other changes to be made that I might have missed, please let me know. 

@al3xa23

PR #997, opened 2026-01-26
Add new delta_sharing node-type

riteshk001:
A new node-type `delta sharing` is implemented to allow the user to fetch, read and write Apache tables using the Delta Sharing protocol. 

#### Features: 
- Fetch metadata from the delta sharing server (query metadata, list schemas, list shares, list tables). 
- Download Apache Parquet tables to a local cache. 
- Requires a profile file to identify the share, schema and table and the bearer token. 

I have used [magpierre/delta-sharing](https://github.com/magpierre/delta-sharing/tree/cppdev/cpp/DeltaSharingClient), which provides C++ based API for accessing Delta Shares natively in C++. 

@al3xa23


riteshk001:
There is a version mismatch of the Apache Arrow packages in Debian and Fedora based package repositories. Debian has the newer version and Fedora 42 has the older version. There are some breaking changes in Apache Parquet because of this, due to which I am installing from source (in deps.sh) for non-Debian based Dockerfiles.

PR #1001, opened 2026-02-09
Parse configuration files using `nlohmann::json`

### LINKED ISSUE #735: Rework configuration parsing
Two main points:

- Drop support for libconfig syntax
- Use YAML parsed

This will keep compatability with JSON-formatted configuration files while introducing YAML as a good human readable format to replace libconfig.

@windrad6 @n-eiling Do we maybe have some students who could work on this?

pjungkamp:
## description

This replaces the `Config::load` functions with an [`nlohmann::json`]-based implementation aiming for correctness and ease of use.

## motivation

The final goal is to port VILLASnode configuration handling top-down to `nlohmann::json`'s idiomatic C++ API. This would then enable us to validate configurations and API requests against our JSON schema using [`json-schema-validator`](https://github.com/pboettch/json-schema-validator/tree/main) inside the `villas-node` process.

This should improve the state of our schemas, as they become a requirement for new features instead of being an afterthought in the development cycle.

## backwards-compatibility

This rewrite *does* change parsing behavior for a few edge cases I considered bugs and isn't feature complete yet.

### include directives

Our include directives currently do not work as documented [here](https://villas.fein-aachen.org/docs/node/config/#node-config-include).

The documented syntax is `{ "@include": "some/glob*pattern" }`. But the actual implementation parses special strings like `"@include some/glob*pattern"` as you can see in e.g. github.com/VILLASframework/node/blob/master/etc/tricks/tricks.json.

I've implemented something mirroring the documented behavior with the difference that I've chosen to use `$include` instead of the libconfig-inspired `@include`, because I consider this to be the more common style for special keys in JSON objects (like `$schema` or `$comment`).

### environment variable substitutions

Our current environment variable substitution code is recursively expanding environment variables and does not specify any way to "escape" our variable substitution syntax.

This means we can make `villas-node` spin indefinitely on recursive substitutions like this:

```bash
ENV_A='${ENV_B}' ENV_B='${ENV_A}' villas-node - << EOF
{ "logging": { "level": "${ENV_A}" } }
EOF
```

And there is no way to have a string that contains `${...}` that does *not* substitute environment variables.

### environment variable substitutions inside include directives

Our current implementation will expand environment variables before globbing without escaping special glob characters. This means that environment variables will be substituted as partial glob expressions. I don't think that this is desirable either. 

### 1. How should we include other files?

As documented, as currently implemented or another way?

### 2. How should we escape special values?

I'm personally in favor of just declaring `$` as special, which then means that *every* literal dollar sign has to be escaped as `$$`.

Another option would be to use `\$` which would need to be doubled to `\\$` since it's also the escape for regular json strings. This would be consistent with the escaping of glob patterns in include directives (These are also escaped using `\` which need to be doubled in JSON).

### 3. Should we escape substituted environment variables in include directives before globbing?

I'm personally not fond of allowing partial patterns to be substituted into the glob pattern. 

---

@stv0g what do you think?

[`nlohmann::json`]: https://github.com/nlohmann/json
[`json-schema-validator`]: https://github.com/pboettch/json-schema-validator


stv0g:
Please rebase onto latest master and make sure GHA passes. Thanks :)

stv0g (approved):

stv0g (approved):

stv0g (dismissed):

pjungkamp:
@stv0g I'm fighting the outdated versions of `libconfig` and `libjansson` on debian...

libconfig 1.7 was released in 2017 and libjansson 2.7 was released in 2021. Both haven't made it into debian bookworm released in 2023.

Just wait until the CI finally passes...

stv0g (approved):

PR #1002, opened 2026-02-10
Cleanup in `common/include/utils.hpp`

pjungkamp:
I've gone through our `utils.hpp` header and removed macros and functions that map directly to the standard library.

Each change is a separate commit and can be dropped if deemed undesirable.

Almost all changes could be done using simple automation like e.g. replacing `ARRAY_LEN` with `std::size`.

The standard library alternatives often provide stronger type safety and prevent subtle bugs.

- Using `ARRAY_LEN` on a decayed pointer to an array would just return 1, while `std::size` only works for non-decayed arrays or library containers.
- `std::swap` will respect ADL `swap` specializations which may elide unneccessary copies.
- `std::min` and `std::max` don't compile if the participating integer types differ. This forces us to highlight narrowing integer conversions with an explicit `static_cast`, instead of an implicit truncation. (We are often mixing `size_t` and `unsigned int` lengths)
- C++20 introduced the `<bit>` header which has functions replacing macros like `IS_POW2`/`LOG2_CEIL`
- Some macros and functions were unused or used only once.

### DIFF
--- a/src/villas-pipe.cpp
+++ b/src/villas-pipe.cpp
@@ -61,7 +61,7 @@ public:
 
     // Initialize memory
     unsigned pool_size =
-        std::bit_ceil(std::max(node->out.vectorize, node->in.vectorize));
+        std::bit_ceil(std::max({node->out.vectorize, node->in.vectorize, 16U}));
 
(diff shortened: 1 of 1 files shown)


stv0g (approved):
Great work! This is really a good example of how a PR should be structured and presented!

A pleasure to review :)

PR #1008, opened 2026-03-25
Fix OpenAPI schema

pjungkamp:
I've noticed several smaller issues with our `doc/openapi` schema.

1. We were using OpenAPI 3.0.3, which isn't fully JSON Schema draft v7 compatible. I've updated our schema to `3.1.0`.
2. The `doc/package.json` and `doc/package-lock.json` bit-rotted away. I did some basic maintenance.
3. We didn't even try to test/lint our schema in CI. I've added a `redocly lint` call to our `build:openapi` CI and fixed or suppressed all errors to make it pass.

---

#### Open Questions

- I've disabled the `security-defined` redocly lint... we don't have any security on our API...
- I've also disabled the lint flagging the `servers` list in `doc/openapi/openapi.yaml` for using `localhost`. Is that list still correct? I can't access the k8s endpoint from the open internet... is there a public endpoint?
- Should we update and maintain the `doc/package.json`, or remove it entirely? It is useful for local development on the schema. 
- Should the `doc/package-lock.json` really be committed into our repository?

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -141,7 +141,8 @@ build:openapi:
   needs: []
   image: node:24-alpine
   script:
-    - npx -y @redocly/cli build-docs doc/openapi/openapi.yaml --output openapi.html
+    - npx -y @redocly/cli lint --config doc/redocly.yaml doc/openapi/openapi.yaml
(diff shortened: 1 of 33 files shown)


stv0g (approved):
Great :) Thanks a lot!

PR #1010, opened 2026-04-13
Fix ci too long job name

al3xa23:
Hi all, 
it seems that long job names are not allowed anymore in the CI. Here is an idea for a fix :)


stv0g (approved):

stv0g:
Thanks for the fix 🥳

PR #1011, opened 2026-04-17
Executable timeseries chronics (villas-chronics)

riteshk001:
@al3xa23  
The implementation for creating chronics data for OpenDSS simulations has been redefined and reworked to function as a standalone executable.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -12,6 +12,24 @@ variables:
   DOCKER_CLI_EXPERIMENTAL: enabled
   CMAKE_BUILD_OPTS: "--parallel 16"
   CMAKE_EXTRA_OPTS: "-DCMAKE_BUILD_TYPE=Release"
+  CMAKE_EXTRA_OPTS_FEDORA_MINIMAL: >
+    -DVILLAS_COMPILE_WARNING_AS_ERROR=ON
+    -DWITH_API=OFF
+    -DWITH_CLIENTS=OFF
+    -DWITH_CONFIG=OFF
+    -DWITH_DOC=OFF
+    -DWITH_FPGA=OFF
+    -DWITH_GRAPHVIZ=OFF
+    -DWITH_HOOKS=OFF
+    -DWITH_LUA=OFF
+    -DWITH_OPENMP=OFF
+    -DWITH_PLUGINS=OFF
+    -DWITH_SRC=OFF
+    -DWITH_TESTS=OFF
+    -DWITH_TOOLS=OFF
+    -DWITH_WEB=OFF
+    -DCMAKE_MODULE_PATH=/usr/local/lib64/cmake
+    -DCMAKE_PREFIX_PATH=/usr/local
   CACHIX_CACHE_NAME: villas
 
 stages:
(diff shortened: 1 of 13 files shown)


packaging/docker/Dockerfile.debian-multiarch
```diff
@@ -152,6 +152,11 @@ RUN apt-get update && \
 	libmodbus5:${ARCH} && \
 	rm -rf /var/lib/apt/lists/*
 
+RUN apt-get update && apt-get install -y \
```
stv0g: Please do not create a new `RUN` statement, but rather include the depdendencies in the existing list above.

packaging/deps.sh
```diff
@@ -602,6 +602,32 @@ if ! cmake --find-package -DNAME=ghc_filesystem -DCOMPILER_ID=GNU -DLANGUAGE=CXX
     popd
 fi
 
+# Build and install nlohmann/josn required for villas-chronics
+if ! pkg-config "nlohmann_json" &&
+    should_build "nlohman_json" "for the delta-sharing node-type"; then
```
stv0g, line 607: Please adjust the description

src/villas-chronics.cpp
```diff
@@ -0,0 +1,481 @@
+/* Create chronics files for OpenDSS
```
stv0g:
```suggestion
/* Create chronics files for OpenDSS.
```

src/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ set(SRCS
     villas-signal
     villas-compare
     villas-test-config
+    villas-chronics
```
stv0g: Please make the compilation optional with a `option()` or `make_dependent_option()` statement. This allows us to compile this command only of the required dependencies are present. Otherwise also minimal VILLASnode installation will require nlohmann json and bzip2.

.gitlab-ci.yml
stv0g: This change was already merged as part of https://github.com/VILLASframework/node/pull/1010

Make sure your PR is targeting the latest master branch.

CMakeLists.txt
```diff
@@ -191,6 +193,7 @@ cmake_dependent_option(WITH_SRC                 "Build executables"
 cmake_dependent_option(WITH_TESTS               "Run tests"                                             "${WITH_DEFAULTS}" "TOPLEVEL_PROJECT" OFF)
 cmake_dependent_option(WITH_TOOLS               "Build auxilary tools"                                  "${WITH_DEFAULTS}" "TOPLEVEL_PROJECT" OFF)
 cmake_dependent_option(WITH_WEB                 "Build with internal webserver"                         "${WITH_DEFAULTS}" "LIBWEBSOCKETS_FOUND" OFF)
+cmake_dependent_option(WITH_VILLAS_CHRONICS     "Build with villas-chronics"                            "${WITH_DEFAULTS}" "BZIP2_FOUND" OFF)
```
stv0g: What about nlohman json here?

stv0g (changes requested):

CMakeLists.txt
riteshk001: I have now included that too.

al3xa23:
Hi @stv0g , what can we do proceed with this PR? Please let us know :)

al3xa23:
@pjungkamp , do you have time to do the code review, please? Otherwise, let us know :)

steffen-vogel-opalrt:
After some discussion between @al3xa23 @pjungkamp and me, we concluded that a simple shell script would be a far simpler approach to do this sort of file conversions.

I have tasked GPT-5.4 mini with rewriting the C++ code to a Bash script in #1041 

Please have a look and check if it fulfills your needs.

Also, I am still struggeling to find public reference / documentation about this file format.

There is the following page from the grid2op project:
https://grid2op.readthedocs.io/en/latest/user/chronics.html#module-grid2op.Chronics

But I have a hard time understanding how this is related to OpenDSS.

I am open to include the shell script to VILLASnode, if it is reusable by possibly other users.
If its a one-off / custom conversion for a specific use-case, I believe we are better of in keeping it out.

riteshk001:
@steffen-vogel-opalrt thank you for also designing the shell script. 
I think you are right, the shell script does the same job. We can skip the executable if @al3xa23 agrees. I shall confirm with @al3xa23 and if it is really not needed, then we can close this PR.

steffen-vogel-opalrt:
Thanks for the feedback. I already discussed this with Alexandra during our last VILLAS workshop. I will close the PR then.

PR #1013, opened 2026-04-24
Apply debug_r_session.patch when installing libiec61850 dependency

Hobbbbes:


windrad6 (approved):

stv0g:
libiec61850 v1.6.2 has fixed this and makes the patch obsolete.

Closing in favour of #1024

PR #1017, opened 2026-06-17
Add new FMU node-type

riteshk001:
@al3xa23 

New node type implementation based on Functional Mock-up Interface (FMI 3.0). Currently supports [Co-Simulation](https://fmi-standard.org/docs/3.0/#fmi-for-co-simulation). Functionality can be further extended to include [Model Exchange](https://fmi-standard.org/docs/3.0/#fmi-for-model-exchange) simulation too.


etc/examples/nodes/fmu.conf
```diff
@@ -0,0 +1,61 @@
+# SPDX-FileCopyrightText: 2014-2023 Institute for Automation of Complex Power Systems, RWTH Aachen University
```
stv0g: Fix copyright years

cmake/FindFMI.cmake
```diff
@@ -0,0 +1,26 @@
+# CMakeLists.txt.
+#
+# Author: Ritesh Karki
```
stv0g: Please add a contact mail address

doc/openapi/components/schemas/config/nodes/_fmu.yaml
```diff
@@ -0,0 +1,7 @@
+# yaml-language-server: $schema=http://json-schema.org/draft-07/schema
+# SPDX-FileCopyrightText: 2014-2023 Institute for Automation of Complex Power Systems, RWTH Aachen University
```
stv0g: Fix copyright years

etc/examples/nodes/fmu.conf
```diff
@@ -0,0 +1,61 @@
...
+        in = {
+            signals = (
+                { name = "sine1",   signal = "sine",   amplitude = 1, frequency = 100   },
+                # { name = "in2",   signal = "constant", amplitude = 1  },
+            )
+        }
+    },
+    fmu_node = {
+        type = "fmu"
+        # Path to fmu file
```
stv0g:
```suggestion
        # Path to the FMU file (.fmu)
```

etc/examples/nodes/fmu.conf
```diff
@@ -0,0 +1,61 @@
...
+        fmu_path = "/workspaces/node/fmu_temp/asine.fmu"
+        fmu_unpack_path = "/workspaces/node/fmu_temp/fmu_asine"
+        fmu_writefirst = true
```
stv0g:
```suggestion
        fmu_write_first = true
```

etc/examples/nodes/fmu.conf
```diff
@@ -0,0 +1,61 @@
...
+        stopTime = 10.0
```
stv0g:
```suggestion
        stop_time = 10.0
```

Please use consistent setting names.

etc/examples/nodes/fmu.conf
```diff
@@ -0,0 +1,61 @@
...
+        startTime = 0.0
+        stepSize = 0.2
+
+        in = {
+            signals = (
+                {name = "In1", type = "float"}
```
stv0g:
```suggestion
                { name = "in1", type = "float" }
```

include/villas/nodes/fmu.hpp
```diff
@@ -0,0 +1,76 @@
+/* Node type: Functional Mock-up unit.
+ *
+ * Author: Ritesh Karki <ritesh.karki@rwth-aachen.de>, Jitpanu Maneeratpongsuk <jitpanu.maneeratpongsuk@rwth-aachen.de>
+ * SPDX-FileCopyrightText: 2025 Institute for Automation of Complex Power Systems, RWTH Aachen University
```
stv0g: Fix copyright years

include/villas/nodes/fmu.hpp
```diff
@@ -0,0 +1,76 @@
+/* Node type: Functional Mock-up unit.
```
stv0g:
```suggestion
/* Node type: Functional Mock-up Unit.
```

include/villas/nodes/fmu.hpp
```diff
@@ -0,0 +1,76 @@
...
+struct Sample;
+
+class FmuNode : public Node {
+protected:
+  int parse(json_t *json) override;
+
+  int _read(struct Sample *smps[], unsigned cnt) override;
+  int _write(struct Sample *smps[], unsigned cnt) override;
+
+  bool writingTurn = true;
```
stv0g: Use consistent variable naming (check VILLAS coding conventions).

include/villas/nodes/fmu.hpp
```diff
@@ -0,0 +1,76 @@
...
+  const char *path;
+  const char *unpack_path;
+
+  timespec ts;
+  pthread_mutex_t mutex;
```
stv0g, line 44: Can we use C++ threads?

lib/nodes/fmu.cpp
```diff
@@ -0,0 +1,353 @@
+/* Node type: Functional Mock-up unit.
+ *
+ * Author: Ritesh Karki <ritesh.karki@rwth-aachen.de>, Jitpanu Maneeratpongsuk <jitpanu.maneeratpongsuk@rwth-aachen.de>
+ * SPDX-FileCopyrightText: 2025 Institute for Automation of Complex Power Systems, RWTH Aachen University
```
stv0g: Fix copyright years

packaging/deps.sh
```diff
@@ -587,6 +587,21 @@ if ! find /usr/{local/,}{lib,bin} -name "libOpenDSSC.so" | grep -q . &&
     echo "${PREFIX}/openDSSC/bin/" > /etc/ld.so.conf.d/opendssc.conf
 fi
 
+# Build & Install fmi-library
+if ! find /usr/{local/,}{lib,lib64} -name "libfmilib_shared.so" | grep -q . &&
+    should_build "fmi-library" "For FMI node-type"; then
+    git clone https://github.com/modelon-community/fmi-library.git
```
stv0g: Pin version or Git ref

packaging/deps.sh
```diff
@@ -587,6 +587,21 @@ if ! find /usr/{local/,}{lib,bin} -name "libOpenDSSC.so" | grep -q . &&
     echo "${PREFIX}/openDSSC/bin/" > /etc/ld.so.conf.d/opendssc.conf
 fi
 
+# Build & Install fmi-library
+if ! find /usr/{local/,}{lib,lib64} -name "libfmilib_shared.so" | grep -q . &&
```
stv0g: Does fmilib provide CMake or pkg-config files?

tests/integration/node-fmu.sh
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env bash
+#
+# Integration test for node fmu.
+#
+# Author: Ritesh Karki <ritesh.karki@rwth-aachen.de>
+# SPDX-FileCopyrightText: 2014-2026 Institute for Automation of Complex Power Systems, RWTH Aachen University
+# SPDX-License-Identifier: Apache-2.0
+
+echo "Test not ready yet"
```
stv0g: Please demonstrate that this node-type is working.

doc/openapi/components/schemas/config/nodes/fmu.yaml
```diff
@@ -0,0 +1,71 @@
...
+  properties:
+    fmu_path:
+      format: uri
+      description: |
+        Specifies the URI to a FMU model
+
+    fmu_unpack_path:
+      format: uri
+      description: |
+        Specifies the URI to a directory which has the unpacked files of the FMU model (mainly the shared library and the modelDescription.xml for the model)
```
stv0g: End description consistently with a period

etc/examples/nodes/fmu.conf
```diff
@@ -0,0 +1,61 @@
+# SPDX-FileCopyrightText: 2014-2023 Institute for Automation of Complex Power Systems, RWTH Aachen University
+# SPDX-License-Identifier: Apache-2.0
+
+nodes = {
+    file_output = {
+        type = "file"
+        # format = "csv"
+        uri = "/workspaces/node/fmu_temp/output.dat"
+        out = {
```
stv0g: Is this empty section required?

lib/nodes/fmu.cpp
```diff
@@ -0,0 +1,353 @@
...
+int FmuNode::prepare() {
+  callbacks.malloc = malloc;
+  callbacks.calloc = calloc;
+  callbacks.realloc = realloc;
+  callbacks.free = free;
+  callbacks.logger = nullptr;
+  callbacks.log_level = jm_log_level_info;
+  callbacks.context = nullptr;
+  context = fmi_import_allocate_context(&callbacks);
+
```
stv0g, line 61: I believe at least the callbacks can be initialized in the constructor?

lib/nodes/fmu.cpp
```diff
@@ -0,0 +1,353 @@
...
+  struct stat sb;
+  if (stat(unpack_path, &sb) != 0)
+    std::cerr << "The unpack path is invalid" << std::endl;
```
stv0g: Dont log via iostreams. Use the existing logging infrastructure.

lib/nodes/fmu.cpp
```diff
@@ -0,0 +1,353 @@
...
+    get_vr(signalIn[i].name.data(), signalIn[i].ref, signalIn[i].type);
+  }
+  for (size_t i = 0; i < signalOut.size(); i++) {
+    get_vr(signalOut[i].name.data(), signalOut[i].ref, signalOut[i].type);
+  }
+
+  return Node::prepare();
+}
+
+int FmuNode::check() { return Node::check(); }
```
stv0g: Code formatting?

lib/nodes/fmu.cpp
```diff
@@ -0,0 +1,353 @@
...
+int FmuNode::_read(struct Sample *smps[], unsigned cnt) {
+  assert(cnt == 1);
+
+  const bool hasInputs = !signalIn.empty();
+  // Lock thread
+  pthread_mutex_lock(&mutex);
+  while (hasInputs && (writingTurn || currentTime >= stopTime)) {
+    pthread_cond_wait(&cv, &mutex);
+  }
+  double targetTime = currentTime + stepSize;
```
stv0g: I think this not executing the FMU using a real-time scale?

Can you elaborate a bit how this node-type is indentend to be used?
E.g. does it expect periodic input data to proceed its internal time?
How does the internal time related to the real time?

Do we already have a PR for the node-type? I think this deserves some explanation.

stv0g (changes requested):
Nice work. Its great to see FMI support landing in VILLAS

packaging/deps.sh
al3xa23: I included to look for cmake file

tests/integration/node-fmu.sh
al3xa23: On the to-do-list :)

lib/nodes/fmu.cpp
al3xa23: Right now, it needs periodic input data. My plan would be to include real-time later

doc/openapi/components/schemas/config/nodes/fmu.yaml
```diff
@@ -0,0 +1,71 @@
...
+      format: uri
+      description: |
+        Specifies the URI to a FMU model
+
+    fmu_unpack_path:
+      format: uri
+      description: |
+        Specifies the URI to a directory which has the unpacked files of the FMU model (mainly the shared library and the modelDescription.xml for the model).
+
+    fmuWritefirst:
```
pjungkamp:
```suggestion
    fmu_write_first:
```

doc/openapi/components/schemas/config/nodes/fmu.yaml
```diff
@@ -0,0 +1,71 @@
...
+      type: boolean
+      description: |
+        Specifies whether the model has to be writtem with input values before running. `true` denotes that the model should use the input values. `false` would make the model use default values instead.
+
+    startTime:
```
pjungkamp:
```suggestion
    start_time:
```

doc/openapi/components/schemas/config/nodes/fmu.yaml
```diff
@@ -0,0 +1,71 @@
...
+      type: number
+      description: |
+        Specifies the start time for the simulation. Value has to be at least one `stepSize` lesser than `stopTime`.
+
+    stopTime:
```
pjungkamp:
```suggestion
    stop_time:
```

doc/openapi/components/schemas/config/nodes/fmu.yaml
```diff
@@ -0,0 +1,71 @@
...
+      type: number
+      description: |
+        Specifies the stop time for the simulation. The number of steps is calculated then by the difference between `startTime` and `stopTime` divided by the `stepSize`.
+
+    stepSize:
```
pjungkamp:
```suggestion
    step_size:
```

etc/examples/nodes/fmu.conf
```diff
@@ -0,0 +1,58 @@
...
+
+        limit = 100
+
+        in = {
+            signals = (
+                { name = "sine1",   signal = "sine",   amplitude = 1, frequency = 100   },
+                # { name = "in2",   signal = "constant", amplitude = 1  },
+            )
+        }
+    },
```
pjungkamp:
```suggestion
```

The example configuration file is embedded into the documentation and should **only** contain the node that this example is for. Runnable examples aren't part of this repo.

etc/examples/nodes/fmu.conf
```diff
@@ -0,0 +1,58 @@
...
+paths = (
+    {
+        in = "signal_node",
+        out = "fmu_node"
+    },
+    {
+        in = "fmu_node",
+        out = "file_output"
+    },
+)
```
pjungkamp:
```suggestion
```

See above.

include/villas/nodes/fmu.hpp
pjungkamp: Preferably `std::jthread` (we're on C++20) which automatically provides a `std::stop_token` for clean termination.

include/villas/nodes/fmu.hpp
```diff
@@ -0,0 +1,77 @@
...
+  pthread_cond_t cv;
+
+  fmi3_import_t *fmu;
+  jm_callbacks callbacks;
+  fmi_import_context_t *context;
+
+public:
+  FmuNode(const uuid_t &id = {}, const std::string &name = "");
+  struct fmu_signal {
```
pjungkamp: Types/classes defined in villas-node are PascalCase. Rename `fmu_signal` to `FmuSignal` and move it outside the `FmuNode`.

lib/nodes/fmu.cpp
```diff
@@ -0,0 +1,367 @@
...
+  json_t *json_signals_out = nullptr;
+  json_error_t err;
+  startTime = 0;
+  stepSize = 0.1;
+  stopTime = INT_MAX;
+  ret = json_unpack_ex(
+      json, &err, 0, "{s:s, s:s, s?:f, s?:f, s?:f, s?:{s:o}, s?:{s:o}}",
+      "fmu_path", &path, "fmu_unpack_path", &unpack_path, "stopTime", &stopTime,
+      "startTime", &startTime, "stepSize", &stepSize, "in", "signals",
+      &json_signals_in, "out", "signals", &json_signals_out);
```
pjungkamp:
```suggestion
  ret = json_unpack_ex(
      json, &err, 0, "{s:s, s:s, s?:f, s?:f, s?:f, s?:{s:o}, s?:{s:o}}",
      "fmu_path", &path, "fmu_unpack_path", &unpack_path, "stop_time", &stopTime,
      "start_time", &startTime, "step_size", &stepSize, "in", "signals",
      &json_signals_in, "out", "signals", &json_signals_out);
```

lib/nodes/fmu.cpp
```diff
@@ -0,0 +1,367 @@
...
+      break;
+    case villas::node::SignalType::BOOLEAN:
+      val = smps[0]->data[i].b;
+      break;
+    default:
+      break;
+    }
+    logger->debug("Input value {}", val);
+    switch (signalIn[i].type) {
+    case fmi3_base_type_enu_t::fmi3_base_type_float64:
```
pjungkamp:
```suggestion
    using enum fmi3_base_type_enu_t;
    case fmi3_base_type_float64:
```

This is just a suggestion. You can introduce a `using enum` statement before the first `case` statement to make the cases a little more readable.

lib/nodes/fmu.cpp
```diff
@@ -0,0 +1,367 @@
...
+  if (ret)
+    throw ConfigError(json, err, "node-config-node-fmu");
+  if (stopTime == INT_MAX) {
+    stopTimeDef = false;
+  } else {
+    stopTimeDef = true;
+  }
+
+  parse_signal(json_signals_in, true);
+  parse_signal(json_signals_out, false);
```
pjungkamp: It seems that you only need the `name` from the `in.signal`/`out.signals`. That name is already parsed as part of `Node::parse` above. You can get the names by iterating the signal lists from `getInputSignals` or `getOutputSignals`, which would be the more idiomatic approach here.

lib/nodes/fmu.cpp
```diff
@@ -0,0 +1,367 @@
...
+  return 0;
+}
+
+// Register node
+static char n[] = "fmu";
+static char d[] = "Interface to Functional Mockup Unit using FMI3.0";
+static NodePlugin<FmuNode, n, d,
+                  (int)NodeFactory::Flags::SUPPORTS_READ |
+                      (int)NodeFactory::Flags::SUPPORTS_WRITE |
+                      (int)NodeFactory::Flags::SUPPORTS_POLL>
```
pjungkamp: Does your node support poll? Seems to me like it doesn't.

pjungkamp (changes requested):
I really like the general coding style, it's clean and very readable. Could you please fix the commit such that you have one commit that introduces the FMU node type. You have two unrelated changes changing the capitalization of `writing_turn` in `opendss.{hpp,cpp}` which should not be part of that commit. You can either move those changes into their commit or drop them.

stv0g:
As per the new Contribution guidelines, could you please briefly state if LLMs have been used for this contribution?

lib/nodes/fmu.cpp
```diff
@@ -0,0 +1,363 @@
...
+  return Node::start();
+}
+
+int FmuNode::stop() {
+  fmi3_import_terminate(fmu);
+  fmi3_import_free_instance(fmu);
+  fmi3_import_destroy_dllfmu(fmu);
+  fmi3_import_free(fmu);
+  fmi_import_free_context(context);
+  return 0;
```
pjungkamp: Isn't this missing a `Node::stop`? Could you add an explaining comment if not.

packaging/deps.sh
```diff
@@ -466,6 +466,7 @@ if ! pkg-config "fmt >= 6.1.2" && \
     should_build "fmt" "for logging" "required"; then
     git clone ${GIT_OPTS} --branch 12.1.0 --recursive https://github.com/fmtlib/fmt.git
     mkdir -p fmt/build
+    git checkout 12.1.0
```
pjungkamp: Isn't this covered by the `--branch` option of the `git clone` operation? Why are youa dding this?

> -b<name>
> --branch=<name>
> 
> Point the newly created HEAD to <name> branch instead of the branch pointed to by the cloned repository’s HEAD. In a non-bare repository, this is the branch that will be checked out. --branch can also take tags and detaches the HEAD at that commit in the resulting repository.

packaging/deps.sh
```diff
@@ -587,6 +588,21 @@ if ! find /usr/{local/,}{lib,bin} -name "libOpenDSSC.so" | grep -q . &&
     echo "${PREFIX}/openDSSC/bin/" > /etc/ld.so.conf.d/opendssc.conf
 fi
 
+# Build & Install fmi-library
+if ! find /usr/{local/,}{lib,lib64} -name "libfmilib_shared.so" -o -name "*FMILIB*.cmake"| grep -q . &&
+    should_build "fmi-library" "For FMI node-type"; then
+    git clone https://github.com/modelon-community/fmi-library.git
```
pjungkamp: Can you add a `--branch` here for the latest release tag you support? We don't want to install an unpinned master branch.

riteshk001:
The main logic was written by Jitpanu before. ~~I modified some parts of the code myself and but I also used an LLM mainly to clean and format the code.~~

Edit: Sorry, I just saw the new Contribution guidelines regarding the use of LLMs, 

- Cursor's LLM was used for formatting and cleaning the code in fmu.cpp. 
- The same tool was also used to reformat the file to ensure variable naming adhered to the coding standards mentioned in the documentation.

riteshk001:
Also, clang-format seems to format functions which have just a single line of body by rewriting it into a single line. I think that is why every time I ran clang-format, it would reformat the function body. I think that's why the code formatting looked different to what you expected. 

For the integration test, I did not know where to place the fmu file in the repository, otherwise the intergration test itself is working. But once I add the fmu file, I can remove the "Test not ready" comment from the integration test.

lib/nodes/fmu.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+int FmuNode::prepare() {
+  callbacks.malloc = malloc;
+  callbacks.calloc = calloc;
+  callbacks.realloc = realloc;
+  callbacks.free = free;
+  callbacks.logger = nullptr;
+  callbacks.log_level = jm_log_level_info;
+  callbacks.context = nullptr;
+  context = fmi_import_allocate_context(&callbacks);
```
pjungkamp: Aren't you overwriting the context structure allocated in the constructor above? This seems to me like a memory leak.

pjungkamp (changes requested):
Just some more smaller nitpicks here.

lib/nodes/fmu.cpp
riteshk001: Yes you're right. I don't know how I didn't notice it, a Node::stop is required here.

pjungkamp:
I'm happy with with the code. Run `clang-format`, squash it, and that should be fine.

## integration test

On the topic of including an FMU in this repository for integration testing:

1. It must be redistributable under ASL2.
2. It must be linked against an old version `glibc` to ensure that it runs on all our supported systems. FMU's compiled on up-to-date distributions tend to have problems running with e.g. the `glibc` of older distributions.
3. I'd just put it next to the node-fmu.sh test script. Files with other file extensions than `.sh` [won't be picked up by our integration test suite anyway](https://github.com/VILLASframework/node/blob/97269d6afa7856b9215d2f22ff03f213d10495ef/tools/integration-tests.sh#L60). @stv0g Do you have a better place in mind?

I see an `asine.fmu` file in your integration test, where did that come from?

## on the topic of binary blobs

We're linking our Nix builds against OpalRT's binary OpalOrchestra blobs, so I think that a binary FMU blob might be fine here, even if I don't like it. Binary blobs for testing are the way that the [infamous `xz` backdoor](https://en.wikipedia.org/wiki/XZ_Utils_backdoor) was introduced into the otherwise open-source code.

I'd still like to see a way to build the FMU in a binary-reproducible fashion from source, but that tends to be rather hard. @stv0g We've recently discussed some better modelica/FMU tooling, do you have any good ideas here? 

## out of curiosity

I've taken a little more time to look into what the FMU actually does and have a few more questions. I don't want these to block the merge, you also don't have to change anything. Just some questions to satisfy my curiosity on some of your choices in the implementation.

1. **fmu_write_first:** Can you think of a reason to use `fmu_write_first = false`? This seems to me like it's using some default values specified inside the FMU description. There's also a `default` field on our signal lists that could be used to provide a default value, but it's not used much in VILLASnode. 

2. **fmi3_import_do_step:** Your `_read` and `_write` implementations are currently running in lock-step, blocking on a condition-variable until the other function has finished its turn. The actual `fmi3_import_do_step` function which calls into the FMU to make progress is called as part of your `_read` implementation. Is there a reason for why you've opted to do this on `_read` rather than `_write`?

3. **node/hook**: It looks to me like the FMU is effectively a processing step. You translate our signals to the FMU inputs, do an FMU step, and read the output signals back out. This looks to me like this _could potentially_ be implemented as a `Hook::process` function without any locking instead of a node-type with `_read` and `_write` running in lock-step.  @stv0g What's your take here?

riteshk001:
The asine.fmu file is a simple fmu model that I exported from a Simulink model of arcsine. I wanted to have a simple model to test the implementation and later include it in the integration test for the node. But there are also reference models that are given by Modelica ([Reference FMUs](https://github.com/modelica/Reference-FMUs)) and licensed under BSD-2.

To answer your questions regarding the implementation: 

1. The main reason I included the fmu_write_first was precisely to use the default values of the FMU model. The default values live inside the FMU model. Using the default field of the signal would potentially override the default values if the user supplied the wrong values, that's why I thought it was better to provide the option of directly using the default values provided by the FMU model.  

2. I also wanted to follow the same style of implementation as done in node opendss. Since the idea was to pass input data to the FMU model whenever the node receives it, I felt it made sense to have it in the _read() call instead of the _write() call. 

3. Is it possible for the Hook::process to have multiple signals coming  in or going out? I may be wrong but it felt easier  to have it implemented as a node because of the flexibility it offered when it comes to having multiple paths. I think having it as a node also makes it possible to expand on the functionality of FMU to have its own simulation time later?

riteshk001:
I am a little confused with regards to variable naming conventions, the guidelines say that camel case is required for naming variables, functions and types. But across many cpp/hpp files, snake case is used instead. Which one should I stick to? :)

pjungkamp:
> I am a little confused with regards to variable naming conventions, the guidelines say that camel case is required for naming variables, functions and types. But across many cpp/hpp files, snake case is used instead. Which one should I stick to? :) 

I know that pain. The old C code is pretty much entirely snake_case, so we've got a ton of free functions that are snake_case. We're also using encouraging use of the C++ standard library which also is entirely snake_case. Other libraries like `fmt` and `spdlog` also follow the style of the standard libaray, so we can't escape it. The C++ code introduced camelCase for member functions and virtual member functions of interfaces have permiated through the entire codebase.

I think that the current compromise is something like:

- local variables: snake_case
- non-static data members: snake_case
- free functions: snake_case
- member functions: mostly camelCase
- types: PascalCase

Don't use nested classes or type aliases unless it's really necessary.

Note that we can't enforce any of this statically. Take e.g. a `std::back_inserter`, if you want it to work with your type you need to have a snake_case `push_back` method. Or take the associated type aliases you should provide for custom iterator types like `iterator_category`.

And of course every dependency adds another naming convention into the mix...

pjungkamp:
> 3. Is it possible for the Hook::process to have multiple signals coming  in or going out? I may be wrong but it felt easier  to have it implemented as a node because of the flexibility it offered when it comes to having multiple paths. I think having it as a node also makes it possible to expand on the functionality of FMU to have its own simulation time later?

The main limitations of a Hook compared to a loopback-style node like you have implemented here are:

-  a hook can only process one sample at a time (no vectorization/batching)
- a hook can only output zero or one samples per input sample
- a hook can not produce an output without input

Using a node instead of a hook lifts these restrictions but forces the user to use 2 paths in VILLASnode, one for the input, one for the output side. Each path is a separate thread and thus adds extra synchronization overhead.

The current capabilities of the FMU node could well have been implemented as a Hook (as far as I can tell) but I don't know whether the expected advancements for e.g. model exchange and proper simulation time support would clash with the restrictions that a Hook imposes.

pjungkamp:
> The asine.fmu file is a simple fmu model that I exported from a Simulink model of arcsine. I wanted to have a simple model to test the implementation and later include it in the integration test for the node. But there are also reference models that are given by Modelica ([Reference FMUs](https://github.com/modelica/Reference-FMUs)) and licensed under BSD-2.

Can you check the MathWorks/Simulink documentation about the licensing for FMUs exported from Simulink? I've never looked those that before.

A Modelica BSD-2 Clause FMU should be compatible with our Apache-2.0 license.

stv0g:
I believe we should go with one of the Modelica Reference FMUs.

The Statespace one could be a good pick:

https://github.com/modelica/Reference-FMUs/tree/main/StateSpace

They are distributed as C Code which we could easily compile into an FMU without depending on external tools like Simulink or OpenModelica.

At the same time we have a good guarantee that they are Standard Compliant.

We could just built them from source or integrate them as a binary blob which we download. I would be open to both approaches.

But I agree with Philipp, that an integration test should be there.

riteshk001:
I wanted to add the Statespace model too, but the outputs happen to be 3 element arrays which as of now are not supported by VILLAS if I am not wrong. 
That's why I used the Dahlquist.fmu from the list of reference FMUs given by Modelica for the integration test, but it looks like the .fmu file itself is failing the license check by CI.

pjungkamp:
Well, an FMU is just a ZIP file and thus can't be annotated using comments. If we assume a file named `example.fmu` you'd have to add an `example.fmu.license` file right next to the original one. This `.license` file can then include the `SPDX-*` annotations.

You can check the [reuse.software](https://reuse.software/faq/#uncommentable-file) website for the specification of how you should annotate uncommentable files.

tests/integration/Dahlquist.fmu.license
pjungkamp, line 1: You don't need C style comments here. Just put the SPDX headers there. Nothing else.

On a related note: Are you the author? I thought the FMU was a reference from Modelica sources.

pjungkamp (dismissed):
You're close! You'll have to run clang-format once more and fix your commit messages. They have a space between the type and the scope of the commit. Use `type(scope): description` instead of `type (scope): description`.

tests/integration/Dahlquist.fmu.license
riteshk001: Oh no, I did not make any modifications to the FMU, I was testing the license locally and forgot to remove it in the end. I have fixed it now. :)

tests/integration/Dahlquist.fmu.license
```diff
@@ -0,0 +1,5 @@
+/* Reference FMU model for integration test of FMU node-type.
+ *
+ * SPDX-FileCopyrightText: 2026, Modelica Association Project "FMI"
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
```
pjungkamp:
```suggestion
SPDX-FileCopyrightText: 2026, Modelica Association Project "FMI"
SPDX-License-Identifier: BSD-2-Clause
```

pjungkamp (changes requested):

pjungkamp (dismissed):
LGTM! @stv0g Can you take a short look?

tests/integration/Dahlquist.fmu.license
pjungkamp: Wow. Automatically generated commit messages from GitHub suggestions don't work with commitlint... Argh... Sorry...

pjungkamp:
You'll have to rebase onto master and rewrite the JSON schema to match the new style because #1050 has just been merged.

PR #1020, opened 2026-07-14
 node shmem: Reset Sample::signals before writing to shared memory

leonardocarreras:
The shmem node writes samples whose `signals` field is a `std::shared_ptr` into the writer's heap. A reader in another process dereferences that pointer while copying the sample and crashes.

Discovered when using two villas-node processes with shmem, it segfaults at the first read.

### DIFF
--- a/lib/shmem.cpp
+++ b/lib/shmem.cpp
@@ -194,6 +194,10 @@ int villas::node::shmem_int_write(struct ShmemInterface *shm,
 
   atomic_fetch_add(&shm->writers, 1);
 
+  // Signals is a heap shared_ptr; a reader in another process can't deref it.
+  for (unsigned i = 0; i < cnt; i++)
+    const_cast<struct Sample *>(smps[i])->signals.reset();
+
   ret =
       queue_signalled_push_many(&shm->write.shared->queue, (void **)smps, cnt);
 


lib/nodes/shmem.cpp
```diff
@@ -176,6 +176,10 @@ int villas::node::shmem_write(NodeCompat *n, struct Sample *const smps[],
   if (copied < avail)
     n->logger->warn("Outgoing pool underrun");
 
+  /* signals is a heap shared_ptr; a reader in another process can't deref it */
```
stv0g: Some coding style nits:

- Please start comments with an upper-case letter.
- Please them a full sentence by ending them with a period.
- Please use C++ style comments

```suggestion
  // Signals is a heap shared_ptr; a reader in another process can't deref it.
```

stv0g (changes requested):
I believe this change is better located in `shmem_int_write`:

https://github.com/leonardocarreras/node/blob/cd0ac2b4577057a3c384f2c2148fa7f78382b374/lib/shmem.cpp#L190

lib/nodes/shmem.cpp
leonardocarreras: Fixed

leonardocarreras:
> I believe this change is better located in `shmem_int_write`:
> 
> https://github.com/leonardocarreras/node/blob/cd0ac2b4577057a3c384f2c2148fa7f78382b374/lib/shmem.cpp#L190

Hi @stv0g, I agree with your suggestion. I moved it there and tested again. Thanks also for the style comments.

stv0g (approved):

PR #1021, opened 2026-07-14
dp hook: reconstruct harmonics at k*f0, not k Hz

leonardocarreras:
The inverse dp transformation treated `fharmonics[k]` as a frequency in Hz, but it is a list for indexes (`k*f0`), same than the direct transformation uses. So each component was rebuilt at k Hz instead of k*f0 making the fundamental to be approx 1 Hz.


steffen-vogel-opalrt (approved):

stv0g (approved):

stv0g:
Thanks, looks good. I am merging, despite the CI being green as this is very simple fix.

PR #1023, opened 2026-07-24
Add new C37.118 Synchrophasor node-type (reloaded)

### LINKED ISSUE #254: Add C37.118 node-type
/cc @jpnorenam 

- Libraries:
  - https://github.com/jpnorenam/Open-C37.118/tree/refactoring
  - https://github.com/pksadiq/pmu
  - https://github.com/ALSETLab/PMU-PDC-StreamSimulator
- Good protocol description: https://www.typhoon-hil.com/documentation/typhoon-hil-software-manual/References/c37_118_protocol.html
- https://github.com/OpenPMU
- https://wiki.wireshark.org/IEEE%20C37.118
- Papers:
  - https://ieeexplore.ieee.org/document/7443664
  - https://ieeexplore.ieee.org/document/8256606
  - https://core.ac.uk/download/pdf/33592679.pdf

pjungkamp:
# Description

- Resolves #254
- Depends on #1001 

New comprehensive node type implementing the C37.118.2-2011 protocol over TCP acting as both client and server. It features near zero-configuration for the client-side and a relatively simple configuration for the server-side.

```json
{
  "nodes": {
    "c37_118": {
      "type": "c37.118",

      "in": {
        "address": "localhost",
        "idcode": 48879
      },

      "out": {
        "address": "localhost",
        "idcode": 48879,
        "data_rate": 10,
        "pmus": [
          {
            "name": "VILLASnode",
            "idcode": 48879,
            "frequency": "frequency_signal0",
            "rocof": "rocof_signal0",
            "phasor": [
              {
                "signal": "phasor_signal0",
                "unit": "volt"
              },
              {
                "signal": "phasor_signal1",
                "unit": "volt"
              },
              {
                "signal": "phasor_signal2",
                "unit": "volt"
              }
            ]
          }
        ]
      }
    }
  }
}
```

### Unimplemented

- [ ] Request configuration 3 messages when available
- [ ] Multiple data streams on the same server.
- [ ] Detect runtime configuration changes
- [ ] Client reconnection on server-side disconnect (depends on configuration change detection)
- [ ] TLS encryption
- [ ] UDP transport

### Mentions

- @windrad6 
- @stv0g

### DIFF
--- a/common/lib/json.cpp
+++ b/common/lib/json.cpp
@@ -1,4 +1,4 @@
-/* Json configuration parsing implementation.
+/* JSON configuration parsing implementation.
  *
  * Author: Philipp Jungkamp <philipp.jungkamp@rwth-aachen.de>
  * SPDX-FileCopyrightText: 2024-2025 Institute for Automation of Complex Power Systems, RWTH Aachen University
@@ -149,7 +149,7 @@ void to_json(Json &json, JanssonPtr const &jansson) {
 
 namespace {
 
-// implementation of deprecated variable substitutions
+// Implement the deprecated variable substitution syntax.
 void expand_substitutions(Json &value, bool resolve_env,
                           fs::path const *include_dir) {
   if (not value.is_string())
@@ -164,20 +164,20 @@ void expand_substitutions(Json &value, bool resolve_env,
   auto do_include = false;
   auto expanded = std::size_t{0};
 
-  // check for legacy @include keyword
+  // Check for deprecated @include directives.
   if (include_dir and string.starts_with(DEPRECATED_INCLUDE_KEYWORD)) {
     do_include = true;
     expanded = DEPRECATED_INCLUDE_KEYWORD.length();
   }
 
-  // legacy environment variable substitution syntax
+  // Check for deprecated environment variable substitutions.
   static auto const env_regex = std::regex(R"--(\$\{([^}]+)\})--");
   enum : std::size_t {
     CAPTURE_ALL = 0,
     CAPTURE_NAME,
   };
 
-  // expand deprecated environment substition syntax
+  // Try to expand environment substitions.
   std::smatch match;
   while (resolve_env and std::regex_search(string.cbegin() + expanded,
                                            string.cend(), match, env_regex)) {
@@ -193,7 +193,7 @@ void expand_substitutions(Json &value, bool resolve_env,
     expanded += match.position() + env_value.length();
   }
 
-  // expand deprecated @include directive
+  // Try to expand @include directives.
   if (do_include) {
     auto pattern =
         std::string_view(string).substr(DEPRECATED_INCLUDE_KEYWORD.length());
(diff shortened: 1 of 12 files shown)


al3xa23 (dismissed):

lib/nodes/c37_118.cpp
```diff
@@ -0,0 +1,2213 @@
...
+    crc ^= quick;
+    quick <<= 7;
+    crc ^= quick;
+  }
+
+  return crc;
+}
+
+template <typename T, std::size_t shift, std::size_t mask>
+class bit_field_proxy {
```
al3xa23, line 120: Class defintion in header?

lib/nodes/c37_118.cpp
```diff
@@ -0,0 +1,2213 @@
...
+    if (sent < 0)
+      throw SystemError("Failed to send C37.118 frame");
+
+    offset += sent;
+  }
+}
+
+bool C37Node::recvAll(int sd, std::span<std::byte> buf) {
+  std::size_t offset = 0;
+  while (offset < buf.size()) {
```
al3xa23, line 1411: Is there a maximal buffer size allowed in c37?

al3xa23:
Nice impelementation, thanks :)
Some general questions:
Do you plan to do integration test?
Are you already preparing the documentation?

lib/nodes/c37_118.cpp
pjungkamp: I don't add classes to headers that don't have to be shared. The header only exists to allow the unit test to verify the protocol implementation.

lib/nodes/c37_118.cpp
pjungkamp: Yes, 65535 bytes, why?

common/lib/json.cpp
```diff
@@ -0,0 +1,379 @@
...
+    paths = utils::glob(pattern_expanded, std::span(hook->include_dir, 1));
+    std::erase_if(paths, [](auto const &path) {
+      auto ec = std::error_code{};
+      return not fs::is_regular_file(path, ec);
+    });
+  } catch (...) {
+  }
+
+  if (paths.empty()) {
+    *error = "include directive did not match any file";
```
al3xa23, line 315: Maybe I did not see it, but is the error somewhere logged for the user?

pjungkamp:
Yes. Documentation is ready, I just didn't open a PR yet. I'm definitely missing an integration test though. I'll add one when I'm back.

lib/nodes/c37_118.cpp
al3xa23: Not sure if it needs to be checked ... But it is quite large, so probably not

common/lib/json.cpp
pjungkamp: Yes. This function is a callback for a C library and thus is not allowed to throw any exceptions. The error here is later read from `config_error_text` if `config_read` failed.

lib/nodes/c37_118.cpp
pjungkamp: This function is essentially unrelated to C37. You give it a mutable span of bytes and `recvAll` fills that entire span with data. The C37 buffer size isn't part of this function's contract.

stv0g (changes requested):
Some minor nit picks about comment style:

- Please start comments with uppercase letter
- Please end comments with a period
- Please try to form sentences
- Please write acronyms in all uppercase: Json -> JSON

stv0g (approved):

PR #1024, opened 2026-07-29
deps(libiec61850): Update to 1.6.2

windrad6:


stv0g:
This unfortunately breaks the Nix build, which still needs to use the Patch atm.

stv0g:
Somehow this also breaks the node-type witha failing integration test:

```
[node] 05:52:01  err path:1           Failed to read samples from node goose(iec61850-8-1)
```

@windrad6

steffen-vogel-opalrt:
I have updated Nixpkgs. Now also the Nix builds use libiec61850 v.1.6.2 making the patch fully obsolete.
The patch has now been removed.

@windrad6 Lets see if the tests are now passing.

steffen-vogel-opalrt:
Unfortunately the RGOOSE integration test is still failing.

@windrad6 Do you have some time to look at it.. Otherwise we would disable RGOOSE for now..

PR #1032, opened 2026-07-29
ci: run fork pull requests via pull_request_target behind a review gate

leonardocarreras:
Fork PRs like #1020 fail from `prepare` onward, because fork runs get a read-only `GITHUB_TOKEN` and cannot push to ghcr.

Forks can now run via `pull_request_target`, which is safe because each run needs approval from a member. Nothing changes for same-repo PRs. Tags are unchanged.

Needs the env `fork-pr` to be created with required reviewers before merge. Sorry @stv0g, I will need to make you work a bit extra because this needs admin rights.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -9,6 +9,9 @@ on:
       docker_image:
         required: true
         type: string
+      ref:
+        required: true
+        type: string
 
 env:
   CMAKE_BUILD_OPTS: --parallel $(nproc)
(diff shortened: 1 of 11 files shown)


stv0g (approved):

stv0g:
Unfortunately #1020 still seems to fail?

leonardocarreras:
> Unfortunately #1020 still seems to fail?

Yes, security in the actions increased, so we need some more explicit flags. Additional fix in #1033, hopefully is the last. At least the environment worked, so I think the rest is correct.

PR #1033, opened 2026-07-30
ci: allow fork PR checkout under pull_request_target

leonardocarreras:
'Fork PRs still fail after #1032. `actions/checkout@v7` refuses to check out fork code from a `pull_request_target` workflow unless `allow-unsafe-pr-checkout: true` is set. Approval by a member is already required by the `fork-pr` env, so this should be safe.

Other smaller fixes
- GitHub refuses Actions cache writes for fork runs, so `cache-to` is disabled for that event. Reads are unchanged.
- Splits the concurrency group by event so the two runs stop cancelling each other.
- Fixes `all`, which was skipped instead of failed when a prerequisite failed. Only the real run claims the required check name now.
- Compares the head repo full name instead of using `head.repo.fork`, which is wrong when the base repo is itself a fork.

Verified using cross-repo PR leonardocarreras/node#2

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -72,6 +72,7 @@ jobs:
         uses: actions/checkout@v7
         with:
           ref: ${{ inputs.ref }}
+          allow-unsafe-pr-checkout: true
           fetch-depth: 0
           submodules: recursive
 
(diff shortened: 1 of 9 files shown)


leonardocarreras:
Hi @stv0g! So, as it got tricky I tested here https://github.com/leonardocarreras/node/actions/runs/30521609262?pr=2

stv0g (approved):

PR #1039, opened 2026-07-31
chore: Add renovate configuration file for dependency management

stv0g:


leonardocarreras (approved):
Looks good to me... 

Options for follow up PRs to increase coverage are:
- Containers
```json
{
  "matchManagers": ["dockerfile"],
  "matchFileNames": ["packaging/docker/**"],
  "matchUpdateTypes": ["major"],
  "enabled": false
}
```
- Python 
```json
{
  "matchManagers": ["pep621"],
  "groupName": "Python dependencies"
}
```

steffen-vogel-opalrt:
I am wondering: why is the config for containers not enabled?

leonardocarreras:
Sorry, my comment was misleading/confusing (does not increase coverage, just helps to define options)... Scan is enabled in the default recommended config...

That one was more on the direction that I understand that the `"enabled": false` would help to avoid it to ask you to move i.e. from fedora 42 -> 44, if you do not want that to be automatic. Another option is `"dependencyDashboardApproval": true` and lets you decide when it happens... But is nothing critical in any case...

stv0g:
I see. That was my assumption as well.. I think it's a good thing that we get PRs also about major updates. Wether we accept or close them is another point though

PR #1040, opened 2026-08-02
Remove GPL-2.0 licensed pcimem tool and related files

steffen-vogel-opalrt:
- Was only used as a debugging helper in VILLASfpga
- Is also still available elsewhere
- Allows us to drop more GPL licensed files from the repo


leonardocarreras (approved):
Looks okay.

We might want to tell the user how/where to look for this, in case that is needed, but that can definitely be a documentation concern.

PR #1041, opened 2026-08-02
Add grid2op-timeseries-converter script for chronics file generation

steffen-vogel-opalrt:
Successor to #1011 

This is an AI-support port of the C++ code in #1011 to a Bash script.
The code has been generated using GPT-5.4 mini

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -25,3 +25,10 @@ install(
     DESTINATION ${CMAKE_INSTALL_BINDIR}
     RENAME villas-api
 )
+
+install(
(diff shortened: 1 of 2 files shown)


pjungkamp (changes requested):
This seems like a more sensible approach for the problem.

A C++ tool is a always a ticking time bomb in bomb in terms of build failures. A simple shell script with tools from the 90s will probably work for the next century.

Did you verify the output against the former C++ version? I'd like to see an integration test, to be sure it's working and won't bit-rot away.

stv0g (approved):

PR #1042, opened 2026-08-02
Update auto-generated code

steffen-vogel-opalrt:
This PR updates the generated Python code from Protobuf descriptions.


python/villas/node/villas_pb2.py
```diff
@@ -22,6 +22,8 @@
 _sym_db = _symbol_database.Default()
 
 
+
+
```
leonardocarreras, line 26: Hi Steffen, I think some commit was not yet pushed, or is this the only diff?

steffen-vogel-opalrt:
Oh sorry. I cherry-picked the change, and apparently missed that it was already applied in master

PR #1043, opened 2026-08-02
Remove superfluous CMake variable

steffen-vogel-opalrt:


pjungkamp (approved):
I didn't notice that we've enabled `EXPORT_COMPILE_COMMANDS` in our top-level `CMakeLists.txt`. LGTM.

PR #1047, opened 2026-08-03
feat(nodes): Add c37.118 node-type with WITH_NODE_C37_118 build option

### LINKED ISSUE #1046: Node compiles unconditionally for C37.118
In https://github.com/VILLASframework/node/pull/1023 there is no WITH_NODE_C37_118 that would allow to activate/deactivate the node like others.  The PR has been merged, so a small fix would help. This could break some builds and changes the behaviour of others.

leonardocarreras:
**Disclaimer:** the title should be (fix) but I deliverately picked (feat) to make release-please to know that the node was there, the actual feat is #1023 (merged), so will get a correct minor bump

_What it actually does here:_

- Create the option for building with/without WITH_NODE_C37_118
- Wire the test and the skip message
- Add checks for netdb and poll, it might need some comments from the original creator @pjungkamp to see if this is the correct way

Closes #1046 and will help with #1045

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,8 @@ add_compile_options(
 check_include_file("sys/eventfd.h" HAS_EVENTFD)
 check_include_file("semaphore.h" HAS_SEMAPHORE)
 check_include_file("sys/mman.h" HAS_MMAN)
+check_include_file("netdb.h" HAS_NETDB)
+check_include_file("poll.h" HAS_POLL)
 
 # Use the switch NO_EVENTFD to deactivate eventfd usage indepentent of availability on OS
 if(${NO_EVENTFD})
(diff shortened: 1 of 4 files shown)


steffen-vogel-opalrt (approved):

steffen-vogel-opalrt:
LGTM

pjungkamp:
I don't agree with the checks for `poll.h` and `netdb.h`. 

`poll.h` is unconditionally required our path implementations which is non-optional. The `netdb.h` header is required for e.g the `socket` and `rtp` node-types (among others).

But even more importantly, why would header checks for POSIX headers even be necessary? Is there any system we want to support where those could be not available? And adding these checks for all POSIX headers (e.g. pthreads) doesn't seem desireable. I'd consider it bloat in our build system.

leonardocarreras:
@pjungkamp thanks for the reply.

Agree on the view regarding the checks.

As a bit of context, I am experimenting with an MSVC build of a minimal non-POSIX subset, but we can deal with that in a different way (maybe `_MSC_VER`). That is still not in-tree, so I could open a PR to delete the checks, if that would be preferable for you.

PR #1049, opened 2026-08-08
A bunch of small LLM-assisted fixes

stv0g:


etc/examples/hooks/skip_first.conf
```diff
@@ -13,7 +13,7 @@ paths = (
                 type = "skip_first"
 
                 seconds = 10
-                # sequence = 10
+                # samples = 10
```
leonardocarreras, line 16: Is this commented on purpose?

lib/path.cpp
```diff
@@ -673,7 +673,8 @@ json_t *Path::toJson() const {
                           json_string(pd->node->getNameShort().c_str()));
 
   json_t *json_path = json_pack(
-      "{ s: s, s: s, s: s, s: b, s: b s: b, s: b, s: b, s: b s: i, s: o, s: o, "
+      "{ s: s, s: s, s: s, s: b, s: b, s: b, s: b, s: b, s: b, s: i, s: o, s: "
+      "o, "
```
leonardocarreras, line 677: maybe would be good to merge 677 with 678

leonardocarreras (commented):
Hi Steffen! I think is generally good. I left comments about cosmetics/intention, but I do not think they are critical, feel free to ignore.

etc/examples/hooks/skip_first.conf
stv0g: Because only one of seconds / samples is supported. Its commented so the alternative configuration is also documented here.

lib/path.cpp
stv0g: Addressed in the last commit

PR #1051, opened 2026-09-02
Hook pmu timing

windrad6:
Updated the pmu hook:
Updated when and on what window exactly the phasor is estimated and how it is timetagged
Added a complex output for the phasor


stv0g:
CI tests are failing

stv0g:
Please also give this PR a meaningful title. Ideally you also assign it the corrects labels.

PR #1052, opened 2026-09-04
Fix sample compare for timestamps

windrad6:
Sample time compare only compared for a positive time diff. Now changed to an abs compare.


windrad6:
Some node integration tests use the signal generator as input and therefor the timestamps are not comparable to the expected values.
Added -T to the compare function in the tests

pjungkamp (approved):

stv0g:
Looks good. Could you explain what motivated this change? Didnt our tests pass before as well?

stv0g (approved):

PR #1062, opened 2026-09-13
fix(api): Restart temp config file needs .json at the end for corret c…

windrad6:
The api uses in the restart call the villas-node.json.XXXXXX schema for temp config files. This is incompatible with the automated restart, as the config parsing does not parse them as JSON config, and therefore villas fails to start.

Changed the schema to villas-node.XXXXXX.json


stv0g (approved):
LGTM

